PR #9, opened 2017-08-30
Fix restart

chrisknewman:
Tracer restarts working. Added temporary write_restart_field_cn, read_restart_field_cn that are simplified, will be removed when bgc restarts are working. Added an optional cmd line arg for namelist, defaults to icepack_in.


chrisknewman:
[icepack_in.txt](https://github.com/CICE-Consortium/Icepack/files/1278346/icepack_in.txt)
[restart_in.txt](https://github.com/CICE-Consortium/Icepack/files/1278347/restart_in.txt)
[run_test.txt](https://github.com/CICE-Consortium/Icepack/files/1278355/run_test.txt)

To test the tracer restart capability do the following:
1. Rename attached files by removing .txt
2. Execute the run_test script

Note that filepaths may need to be changed in run_test, and a gold directory may need to be created.

eclare108213:
Tested the restart is BFB using 'cmp iced.1982-'... for the current icepack_in configuration.  We will want to use a different configuration as the default, but we can fix this in subsequent PRs when we have the new scripts in place.

PR #11, opened 2017-09-15
Add initial version of icepack scripts

apcraig:
based on the CICE scripts.  To use,

  ./icepack.create.case -c ip01 -m cheyenne
or 
  ./icepack.create.case -t smoke -m cheyenne -testid ip06

The restart test is not supported yet.


apcraig:
Sorry about the problems with the mac.  Lets try to work through that quickly.  I will have a look tomorrow morning.  I unilaterally decided it was time to get rid of the old scripts, hope that wasn't premature.

apcraig:
The latest commit should fix the build output issue and address some other features that were updated in CICE today.   Note that the machine file must now have a variable ICE_MACHINE_MAKE defined.  Generally, this will be "gmake".

eclare108213:
With help from @chrisknewman on the Macros file, this worked for me with just a couple of changes:  I needed to
1.  edit icepack.settings to change ICE_IOTYPE from netcdf to none
2.  add ./ in front of the icepack.run command in icepack.submit
I'm not sure if we should change the scripts for either one of these, but it would be nice for them to work without tinkering, if possible.

I will add Macros.loft, env.loft and modify icepack.batch.csh to add the loft machine after this PR is executed.

apcraig:
I'm glad it's working on the mac and wolf.  I just checked in 3 minor changes including fixing the two issues pointed out above.  I'm happy to continue to iterate and to take our time.  When you think it's ready, please approve the PR, and I will push to master.

eclare108213:
This works.  The test cases have not been modified for Icepack, and that should be put in a separate pull request.  Will merge this one.

PR #13, opened 2017-09-20
Nrldata

eclare108213:
Initial commit of forcing module and changes needed to read and run with the ascii column forcing data from NRL.  For now, this version assumes the model time step is the same as the forcing data increment (1 hour) and no time interpolation is available yet.  The forcing fields are applied to all grid cells.

Importantly, I changed the icepack_in configuration, so the answers will not be the same using the default data settings as in the current version.  There were issues that I fixed along the way so I wouldn't be surprised if the answers do change even with the same namelist settings.  Since we're just now getting an initial configuration set up, I'm not worried about it.  I made no attempt to see if this forcing produces reasonably physical results, only that the numbers are in the ballpark and there aren't NaN's in the diagnostic output.

Big changes:
- added icepack_drv_forcing.F90
- added new namelist flags to forcing_nml
- changed some namelist flags for the configuration needed to run with the NRL data
- added calls to new forcing routines from icepack_drv_InitMod.F90 and icepack_drv_RunMod.F90

### DIFF
--- a/configuration/driver/icepack_drv_InitMod.F90
+++ b/configuration/driver/icepack_drv_InitMod.F90
@@ -35,8 +35,7 @@
       use icepack_drv_constants, only: nu_diag
       use icepack_drv_flux, only: init_coupler_flux, init_history_therm, &
           init_history_dyn, init_flux_atm_ocn
-!      use icepack_drv_forcing, only: init_forcing_ocn, init_forcing_atmo, &
-!          get_forcing_atmo, get_forcing_ocn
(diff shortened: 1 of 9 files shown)


eclare108213:
The current icepack_in uses this data set, but the default in the code (i.e. without a namelist setting) is to use constant forcing values, not this data.  We haven't decided whether this data will be the default setting in an Icepack release.  @duvivier has added the data set to the ftp site at NCAR, I believe, and for now we can send users there to get it rather than adding it to the repo.

PR #16, opened 2017-09-26
PR template

duvivier:
Adding pull request template for Icepack repo as per discussion with Elizabeth. Matches that in CICE repo. Second attempt with a clean branch from my forked repo.
Developer(s): Alice DuVivier
Results (bit for bit, roundoff, climate changing): bit for bit
Code review:
Other Relevant Details:


eclare108213:
For the Icepack template, we also need to know whether changes affect the interface with host sea ice models, and in particular whether changes are necessary in those models.   Such changes should be made in CICE with a corresponding PR, so I guess the CICE template also needs to ask whether there are changes needed in Icepack too.

PR #17, opened 2017-09-27
Initial Icepack test suite

eclare108213:
Defining an initial set of tests for an Icepack test suite.
Developer(s):  Elizabeth, following in the footsteps of Matt and Tony
Updated documentation (Y/N):  not yet
Results (bit for bit, roundoff, climate changing):  BFB
Code review: 
Other Relevant Details:  This suite of tests partly passes, but fails on debug configurations and restarts.  I'll work on the debug issue next.  @apcraig @mattdturner  could either of you get the restart tests working?  Icepack does not use the ice.restart_file pointer as in CICE, and instead looks at the ice_ic string in the namelist file icepack_in.  (We can put that back into the Icepack driver if necessary)

### DIFF
--- a/configuration/driver/icepack_drv_init.F90
+++ b/configuration/driver/icepack_drv_init.F90
@@ -625,7 +625,7 @@
              endif
          endif
 
-         nt_aero = max_ntrcr
+         nt_aero = max_ntrcr - 4*n_aero
          if (tr_aero) then
              nt_aero = ntrcr + 1
              ntrcr = ntrcr + 4*n_aero ! 4 dEdd layers, n_aero species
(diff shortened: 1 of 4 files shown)


apcraig:
With regard to exact restart, my understanding was that icepack was not restarting exactly, so I have not tried to formally implement the test.  But I guess it makes sense to get the automated test working even if the test is failing ... so it can be tested.  I will take care of that task.

eclare108213:
I checked that the debug tests pass on pinto.  The issue on the loft machine is with underflows; I usually set a flush-to-zero flag but I don't see how to do that with gfortran.

In fact, the entire test suite passes on pinto, including the restart tests, even though they are not configured correctly.  Beware!

PR #18, opened 2017-09-29
Merge apcraig refactor1 branch

apcraig:
Move most icepack_intfc subroutines to other files, update the exact restart test, fix a couple scripts bugs, rename ice_diag output files.

Developer(s): tcraig
Results (bit for bit, roundoff, climate changing): bit-for-bit
Code review: 
Other Relevant Details:

Move most of the routines in icepack_intfc to other files.  The icepack_intfc still serves as the module for top level routines to use and all the usable public routines are exposed in that file.

Update the make clean so it removes the executable.

Update the exact restart test, writes monthly restarts and runs 14 months then 12 months from month 2 restarts.   

Fix a couple minor bugs in scripts.

Update the ice_diag output file names, append the name of the gridcell instead of a number.

Update the configuration/scripts/options files to be consistent with icepack namelist.


apcraig:
I am not convinced that having "restart_dir" prepended to the ice_ic name is the right way to go.  I imagine there are cases where ice_ic might be some file sitting somewhere else on the file system, unrelated to restart, that you want to use to start a run.  Or maybe it's sitting in the run directory and not the restart directory, or whatever.  While the ice_ic can be the restart file, it isn't always.  Are you sure you want to change that?  And I assume what you want is for the restart_dir to be prepended to ice_ic in the source code so it's actually hardwired there?  Or do you want that to happen in the scripts?  In the scripts it would be harder to do and would probably require some new env settings.

I also found that my 10 day restart test was passing.  When I changed to one year, it failed.  So I think that's consistent with the inconsistency others are seeing.  I was using the new forcing though.

eclare108213:
Okay, that's fine, we can have the path directly in ice_ic.  
I'm surprised that restarts would work with the new forcing, but maybe we've gotten lucky with that.

PR #20, opened 2017-10-07
Climatological forcing and general debugging

eclare108213:
Adding climatological forcing, which requires interpolation, and debugging/cleaning up the forcing module.  I still consider this part of the code to be in process - there will likely be more answer-changes before it settles down.
Developer(s): E. Hunke, Mike Winton
Updated documentation (Y/N): not yet
Results (bit for bit, roundoff, climate changing):  changes answers - fixed lots of bugs in the forcing and also set atm_data_type = 'clim' in icepack_in (was GOFS)
Code review: 
Other Relevant Details:
- Many bugs associated with ocean forcing, including initialization, missing land mask that was resetting everything to zero, sss set to sst value, qdp not initialized.  Note qdp is set in climatological atmo forcing, which is inconsistent and should be generalized at some point.
- Switched GOFS to new data file that includes wind instead of latent and sensible heat; not debugged.
- Added climatological forcing provided by M. Winton, GFDL.
- Cleaned up "unmatched quotes" in comments that gfortran complains about
- subroutine interp_coeff_monthly is used; also added general interp_coeff routine but it is not yet called.
- timeseries.csh no longer tries to plot both Arctic and Antarctic data, which is not available in Icepack's output data files.  However, I did not figure out how to fix the x axis (currently hardwired for hourly output - the default is daily output), or the legend.   @mattdturner do you know how to fix this?

### DIFF
--- a/columnphysics/icepack_mechred.F90
+++ b/columnphysics/icepack_mechred.F90
@@ -1604,7 +1604,7 @@
          dh2rdg     ! change in mean value of h^2 per unit area
                     ! consumed by ridging 
 
-      if (kstrength == 1) then  ! Rothrock '75 formulation
+      if (kstrength == 1) then  ! Rothrock 1975 formulation
 
       !-----------------------------------------------------------------
       ! Compute thickness distribution of ridging and ridged ice.
--- a/columnphysics/icepack_orbital.F90
+++ b/columnphysics/icepack_orbital.F90
@@ -39,7 +39,7 @@
       character (len=*), intent(out) :: stop_label
 
       l_stop = .false.      ! initialized for CCSMCOUPLED
-      stop_label = ''       ! initialized for CCSMCOUPLED
+      stop_label = ' '      ! initialized for CCSMCOUPLED
       iyear_AD  = 1950
       log_print = .false.   ! if true, write out orbital parameters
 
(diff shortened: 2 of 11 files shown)


mattdturner:
I can look into having the timerseries.csh script detect the output timestep (hourly, daily, etc.) and automatically choose the correct timestep for the x-axis.

eclare108213:
Thanks, @mattdturner 
The model does not necessarily write output on every time step.  There are a couple of flags in the namelist that are used to determine when diagnostics are written:  the timestep dt and the diagnostic frequency diagfreq.  E.g. dt=3600 and diagfreq=24 means daily output -- this is the default.  Both Icepack and CICE are set up this way currently.  (@apcraig has suggested that we make this more user-friendly, but we have a lot of other things to do first.)

Also, how can we specify the ranges for the x and y axes?  This doesn't need to be a command-line option -- I'd be fine with simply setting them in the script.  A good starting place: area fraction 0 to 1, ice thickness 0 to 5 m, snow thickness 0 to 0.5 m.  Thank you!

eclare108213:
Discussed this version with @chrisknewman and we agreed that it was ready to merge, so that he can merge his changes into it.

PR #32, opened 2017-11-27
Change GOFS to CFS

dabail10:
Change GOFS to CFS in driver code.
Developer(s): D. Bailey
Updated documentation (Y/N): N
Results (bit for bit, roundoff, climate changing): bfb
Code review: E. Hunke
Other Relevant Details:

PR #33, opened 2017-11-27
Add back in restore_ocn and trestore flags to icepack_in.

dabail10:
Add back in restore_ocn and trestore flags to icepack_in.
Developer(s): D. Bailey
Results (bit for bit, roundoff, climate changing): bfb
Code review: E. Hunke
Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_parameters.F90
+++ b/columnphysics/icepack_parameters.F90
@@ -69,8 +69,8 @@
 !-----------------------------------------------------------------------
 
       character (len=char_len), public :: &
-         shortwave, & ! shortwave method, 'default' ('ccsm3') or 'dEdd'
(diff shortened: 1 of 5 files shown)


apcraig:
This seems to include PR #32, should we delete PR #32?  Would it be easier to just work on a single branch for this set of changes and continue to push updates on that working branch?

dabail10:
I assumed that we would still do it in stages. So accept PR #32 and then do #33. I will use the same branch in the future.

apcraig:
either way is ok, just checking.

apcraig:
OK, after merging #32, this now has a conflict.  Not sure why.  Dave, can you check.

dabail10:
Ok. There was a commit made where my branch was not up to date with master. I have resynced with master and am issuing a new PR.

PR #41, opened 2017-12-05
refactor5

apcraig:
next icepack refactor

Developer(s): tcraig
Results (bit for bit, roundoff, climate changing): bit for bit
Code review: 
Other Relevant Details:

modify bgcISPOL nml setting, turn off tr_bgc_Fe temporarily to be consistent with prior results
modify base_suite list of tests.
add calls in icedrv to write icepack output and check abort status on each icepack interface call.
fix errors associated with missing end of file carriage return in set_* option files.
get rid of use of icepack_parameters in the driver code, now done via query/init methods.
move some constants from icepack_parameters to icepack_constants.
move all use statements to the top of the module in columnphysics.
add max	tracer sizes to icepack_intfc to explicitly allow array allocation	in the driver.
get rid of use of icepack_tracers, all reference now via query/init interfaces.


apcraig:
This is not ready for merging to the master yet, still want to do more, but creating a place holder and providing visibility of the upcoming changes.

apcraig:
I think this is ready for a merge to master now.

PR #42, opened 2017-12-06
Icepack ctest submit failure

mattdturner:
Add a backup option for submitting the CTest results to CDash
Developer(s): Matt Turner
Code review: 
Other Relevant Details:

### DIFF
--- a/configuration/scripts/icepack.run.suite
+++ b/configuration/scripts/icepack.run.suite
@@ -182,7 +182,7 @@ cd base_suite.${testid}
 
 ./results.csh
 
-ctest -S steer.cmake
+./run_ctest.csh
 
 cd ..
 
--- a/configuration/scripts/tests/CTest/run_ctest.csh
+++ b/configuration/scripts/tests/CTest/run_ctest.csh
@@ -1,4 +1,4 @@
-#!/usr/bin/csh
+#!/bin/csh -f
 
 set initargv = ( $argv[*] )
 
--- a/configuration/scripts/tests/CTest/run_ctest.csh
+++ b/configuration/scripts/tests/CTest/run_ctest.csh
@@ -21,7 +21,7 @@ if ( $submit_only == 0 ) then
   ctest -S steer.cmake
 else
   # Find the filename to submit to CDash
-  set CTEST_TAG="`head -n 1 Testing/TAG`"
+  set CTEST_TAG="`head -n 1 Testing/TAG.submit`"
   
 cat > submit.cmake << EOF0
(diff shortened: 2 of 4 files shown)


eclare108213:
This isn't working for me. I created a control suite and then used icepack.run.suite to run the tests, which completed as expected (the usual 2 tests failed). But the run_ctest.csh command either isn't present or isn't recognized as a command (see below). I'm not sure which -- I didn't find it in the local suite directory, so I copied it there by hand, but even then it's not recognized as a command. Also, I'm not sure which directory I should be looking in; I assume Icepack.ctst for this:
[eclare@pi-fe1 icepack_ctest_submit_failure]$ configuration/scripts/icepack.run.suite -m pinto -testid ctst -bc e.fork2

[eclare@pi-fe1 Icepack.ctst]$ ./run_ctest.csh
./run_ctest.csh: Command not found.
[eclare@pi-fe1 Icepack.ctst]$ ls -l run_ctest.csh
-rwxrwxr-x 1 eclare eclare 1942 Dec 7 21:41 run_ctest.csh*
[eclare@pi-fe1 Icepack.ctst]$ module list
Currently Loaded Modulefiles:

 1)   git/2.14.1
 2)   intel/17.0.4 6) use.own
 3)   openmpi/2.1.2
 4)   totalview/2017.0.12
 5) python/3.5-anaconda-4.1.1
 6) use.own
 7) idl/8.5.1

mattdturner:
I believe that this might be due to the fact that icepack.run.suite creates a fresh clone of the CICE-Consortium branch, which does not have contain the run_ctest.csh script.  To test this new implementation prior to it being merged, you can edit icepack.run.suite by replacing:
```
git clone https://github.com/CICE-Consortium/Icepack.git $gitDir

cd $gitDir
```
with 
```
git clone https://github.com/mattdturner/Icepack.git $gitDir

cd $gitDir

git checkout icepack_ctest_submit_failure
```

Once this pull request is merged (after it is confirmed to work), then the modification to icepack.run.suite will no longer be required.

run_ctest.csh should be copied from configuration/scripts/tests/CTest to Icepack.ctst/base_suite.ctst in your case.

eclare108213:
I needed to change the first line of run_ctest.csh from
#!/usr/bin/csh
to
#!/bin/csh -f
Please fix that.

I think it makes sense to move configuration/scripts/icepack.run.suite into the main directory, next to icepack.create.case.  @apcraig do you agree?  I think it would also make sense to call it icepack.create.suite instead of .run. since it also creates directories and builds the code.

I have another set of tests waiting in the queue, using icepack.run.suite modified as above.

PR #43, opened 2017-12-06
Option to pass account number to icepack.create.case

mattdturner:
This pull request adds a '-a' option to icepack.create.case to allow users to pass the queue manager account number as an optional argument.
Developer(s): Matt Turner
Code review: 
Other Relevant Details:

The way this works is by having icepack.create.case write the account number (if it is passed as an argument) to ~/.cice.proj.  The env.<machine> file then reads the account number from ~/.cice.proj.  I have no knowledge of what ~/.cice.proj is used for, so this pull request should be rejected if these edits might break some other functionality.

### DIFF
--- a/doc/source/icepack_3_user_guide.rst
+++ b/doc/source/icepack_3_user_guide.rst
@@ -416,6 +416,8 @@ icepack.create.case generates a case. Use ``create.case -h`` for help with the t
 
   -m is the machine name (required). Currently, there are working ports for NCAR yellowstone and cheyenne, AFRL thunder, NavyDSRC gordon and conrad, and LANL’s wolf machines.
 
+  -a is the account number for the queue manager (default is defined in env.<machine>)
+
   -s are comma separated optional env or namelist settings (default is 'null')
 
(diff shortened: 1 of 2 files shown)


apcraig:
I don't see a problem with this.  The idea behind ~/.cice_proj is for each user to be able to manually set that file and for the account number in there to be used thereafter as their default.  

I guess we should think about whether we want to have an option on the command line to do that for users.  And if .cice_proj already exists, what should the implementation do?  Another idea would be for -a to define the account number just for the current case, overriding what is in the defaults and in .cice_proj.  I think I prefer that method more.

eclare108213:
I agree.  I have more than one project account and prefer to always have one of them as the default.  I wouldn’t want to have to remember (or check) what the last usage was, or to always have to specify -a, but it would be nice to be able to use a different account on the fly, occasionally.

mattdturner:
That makes sense.  I'll go ahead and cancel this PR and work on a separate implementation that doesn't use the ~/.cice_proj file

PR #44, opened 2017-12-07
Icepack.create.case option to pass account number

mattdturner:
Add option to pass account number to Icepack create.cast
Developer(s): Matt Turner
Code review: 
Other Relevant Details:

As it is currently setup, the priority for defining the account number is as follows:
    default defined in env.<machine> < ~/.cice_proj < icepack.create.case -a
In other words, passing an account number to icepack.create.case via the '-a' option overwrites all others.  These edits to env.<machine> are made in the case directory only.

### DIFF
--- a/configuration/scripts/icepack.batch.csh
+++ b/configuration/scripts/icepack.batch.csh
@@ -14,7 +14,7 @@ set jobfile = $1
 set ntasks = ${ICE_NTASKS}
 set nthrds = ${ICE_NTHRDS}
 set maxtpn = ${ICE_MACHINE_TPNODE}
-set acct   = ${ICE_MACHINE_ACCT}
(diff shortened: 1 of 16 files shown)


mattdturner:
@apcraig @eclare108213 
This implementation should be more acceptable.  It no longer has an effect on ~/.cice_proj, aside from to have the -a value supersede the value in ~/.cice_proj

mattdturner:
I've added a new ICE_ACCOUNT variable to icepack.settings.  icepack.create.case now adds the definition of ICE_ACCOUNT to fsmods.  I have also removed all the logic that checks for ~/.cice_proj out of env.<machine> (it is now in icepack.create.case)

eclare108213:
Works for me.

PR #52, opened 2017-12-21
Update documentation, primarily User and Developer Sections

apcraig:
Update Icepack documentation, primarily user and developer sections
Developer(s): tcraig
Are the code changes bit for bit, different at roundoff level, or more substantial? bit-for-bit
Other Relevant Details:

Rearrange and update primarily the user guide section of the rst files.  Includes adding a 
Developer guide section.  Also tried to clean up redundant documentation.

### DIFF
--- a/doc/source/icepack_1_introduction.rst
+++ b/doc/source/icepack_1_introduction.rst
@@ -12,13 +12,19 @@ thickness, enthalpy, ice age, first-year ice area, deformed ice area and volume,
 melt ponds, and biogeochemistry.
 
 Icepack is implemented in CICE as a git submodule.
+Icepack basically consists of three independent parts, the column physics code,
+the icepack driver that supports stand-alone testing of the column physics code, and the
+icepack scripts that build and test the Icepack model.  
+The intent of Icepack is to provide the column physics model as a separate
+library for use in other models such as CICE.
 Development and testing of CICE and Icepack may be done together,
-but development within the repositories is independent.
+but the repositories are independent.
 
 This document uses the following text conventions:
 Variable names used in the code are ``typewritten``.
 Subroutine names are given in *italic*.
 File and directory names are in **boldface**.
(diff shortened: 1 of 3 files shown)


apcraig:
I have done some rearranging of the documentation in Section 3 including adding a Section 4, Developers Guide.  We might want to think about splitting that into a new file, but I didn't with this commit.  Or we might want to merge Section 3 (User) and Section 4 (Developers) back together, although I think I prefer they remain separate.  I have cleaned up some redundant documentation and added some new content.  I struggled a bit with formatting, but think what is there now is OK.  I am happy to rethink/redo that part.  The main problem I had was finding formats that were not double spacing code and scripts.  I didn't like using lists in all cases, I didn't like that line starting with | could not be indented, so I went with literal boxes.  The other problem with the format is inability to manage whitespace, but that's a separate challenge.

This is my first time working with the rst files.  I wonder why we don't include the built html with the master.  It doesn't take that much space, especially if we can find a way to have a single copy of the figures.  I don't see the benefit that we get with using the separate branch to store the html.  I built the html as I was working on the documentation.  I assume others would have to do the same thing to make sure the rst files were fine.  So the html can be there pretty easily and it might be helpful to users to be able to pull up the local html at times.  Even if the html is not generated by developers, it can be generated within the consortium (including updating the pdf file) periodically if the html/pdf falls behind the rst.  Why have the extra branch?  Is it because github can't point to the documentation on master?

dabail10:
The whole rational for a gh-pages orphan branch is that you can just point to:

https://cice-consortium.github.io/Icepack/

for the documentation. If it is under master, then you need to add the full path.

duvivier:
The other reason for using gh-pages instead of master, is based on advice from Alice B. in NCAR cseg based on CIME experience and that this way avoided issues with writing over html that they'd run into. 

I'm working on this, but rusty on my githubbing. I don't think I've resolved conflicts before, but I'll try to get this finished ASAP.

Issue #46, opened 2017-12-07
bgc exact restart fails during icepack testing


apcraig:
PASS cheyenne_restart_col_1x1_bgcISPOL run-initial
FAIL cheyenne_restart_col_1x1_bgcISPOL run-restart

eclare108213:
@mattdturner 
This case is failing because the restart test is looking for restart files in a restart/ directory, which is empty; there are restart files in restart_baseline/.  I don't understand why this case (ISPOL) fails but others do not.  Other cases have restart files in both places.  Can you explain what is going on here, or how it is supposed to work?  Thx

eclare108213:
I suspect that the problem with this case has to do with the start and end times of the ISPOL data.  Since it's from an actual field campaign, the start date is not Jan 1 and it doesn't run for a full year.  

       year_init       = 2004
       istep0          = 4008
       npt             = 4776

The first restart file written is

       ./restart/iced.2004-07-01-00000

and the original test is supposed to only run for 4776 hours (199 days).  The restart test is looking for

     ./restart/iced.2015-03-01-00000

so it looks like the calendar calculation in the restart testing script is not working correctly.

mattdturner:
> This case is failing because the restart test is looking for restart files in a restart/ directory, which is empty; there are restart files in restart_baseline/. I don't understand why this case (ISPOL) fails but others do not. Other cases have restart files in both places. Can you explain what is going on here, or how it is supposed to work? Thx

The way the restart tests work is that the 'run-initial' simulation outputs to the restart/ directory.  The test_restart.script file then renames the restart/ directory to restart_baseline/, copies a restart file to restart/, and runs the 'run-restart' case.

It appears that the restart file date-time group is currently hard-coded to be 2015-03-01.  I'll start looking at a better way to determine a restart date.

eclare108213:
@njeffery @chrisknewman 
I found the bug preventing exact restarts for both ISPOL and NICE bgc cases:  the sst data was read in but not used.  I am running a test suite now to confirm that this fixes the problem and does not create others.

njeffery:
Awesome!


Just a heads up. I found the source of a segmentation fault in algal_dyn when using iron.  I'll let you know when I've implemented a fix.


Nicole

________________________________

eclare108213:
Fixed with PR #60

PR #60, opened 2017-12-27
Restartcleanup

eclare108213:
Began cleaning up restart code, and partially debugged the ISPOL bgc restart problem.
Developer(s):
Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB except ISPOL
Other Relevant Details:
I'll probably add more commits to this PR, but getting it out there for @njeffery or @chrisknewman to look at.  ISPOL case still does not restart exactly.  FIXED


eclare108213:
ISPOL restarts are now exact.  ISPOL sst data was not being used, and so the results change for ISPOL cases.  This addresses issue #46

Issue #59, opened 2017-12-26
creating directories


eclare108213:
When scripts create new directories, please have them first check whether the directory exists.  This example came up when I tried to re-run a restart test in a previously used run directory, using a slightly different namelist configuration.

mkdir: cannot create directory `/net/scratch3/eclare/ICEPACK_RUNS/pinto_restart_col_1x1_debug.testrestart/diags_baseline': File exists

Let's keep an eye out for cases when we really do not want to re-use an existing directory, but most of the time it's probably okay for those of us doing manual shortcuts in/around the usually automated test suite.

mattdturner:
What is the desired behavior?  Do you want the scripts to error out if the directory already exists, or just skip the `mkdir` command and continue?

eclare108213:
Just skip and continue.  This is what we did in the old cice scripts:
if !(-d $EXEDIR) mkdir -p $EXEDIR

PR #61, opened 2017-12-27
No error when creating directories

mattdturner:
Test scripts no longer print errors if a directory already exists.  This resolves https://github.com/CICE-Consortium/Icepack/issues/59

Developer(s): Matt Turner
Other Relevant Details:


mattdturner:
I did not setup the scripts to check if the directory already exists, since `mkdir -p` suppresses errors if the directory already exists.

eclare108213:
Thanks Matt!

PR #83, opened 2018-01-04
Refactor6

apcraig:
Clean up "use" of icepack_constants using instead query function as much as reasonable.

Developer(s): tcraig
Are the code changes bit for bit, different at roundoff level, or more substantial? bit-for-bit
Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_constants.F90
+++ b/columnphysics/icepack_constants.F90
@@ -26,9 +26,8 @@
 
       integer (kind=int_kind), parameter, public :: & 
          nspint = 3                ! number of solar spectral intervals
-                    
+
       real (kind=dbl_kind), parameter, public :: &
-         secday = 86400.0_dbl_kind ,&! seconds in calendar day
          c0   = 0.0_dbl_kind, &
          c1   = 1.0_dbl_kind, &
          c1p5 = 1.5_dbl_kind, &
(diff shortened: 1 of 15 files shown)


apcraig:
I decided to use constants that were parameters and query the others.  In fact, we can go to all-query and maybe should for completeness.  I agree that secday should probably not be a parameter and should then not be used.  I can fix this.  Do you want me to migrate all constants (including true parameters like c0) to query only?

eclare108213:
No, I'd rather we just "use" things that are true numbers and shouldn't be available for users to change.  Anything with units should be queried, and unitless numbers that are nevertheless somewhat arbitrary, like puny.  Does that rule cover everything?

eclare108213:
On the other hand...
If we don't use any constants, then our external code developers would be less likely to avoid coding the query calls by using their newly implemented parameters/constants.  It might be less confusing and easier to deal with, overall.

apcraig:
Can you clarify what you are suggesting with the "other hand".  What does "if we don't use any constants" mean?

eclare108213:
Sorry, I mean "use" as in Fortran "use".  If the only apparent option in the code is to query for things, then developers would be less likely to implement their code with use statements instead of queries, for tuning parameters and such.

apcraig:
I agree.  The other option is that we can easily duplicate the hardcoded parameters, c0, c1, p1, etc in the icedrv_constants directly.  Yes, we would have two copies of the same parameters, but these are well defined so it should be OK.  In fact, maybe that is the right way to do it.  Sharing those constants across the icepack interfaces doesn't really make a lot of sense.

eclare108213:
Yes, and host models other than cice will probably have different names for things like that, e.g. zero, half, one, etc, so it makes even more sense from their point of view to define those in both places.

apcraig:
I just commited a change to the constant parameters and secday.  There is a copy of the constant parameters in the driver now.  I also made secday a non-parameter in the columnphysics and added secday to icedrv_calendar as a public scalar for the driver.  A query during initialization sets the secday in the driver.

eclare108213:
This works for me.  Let's make sure to include the philosophy behind the constants/parameters/tracer init/query routines in the documentation, i.e. that the host model "owns" the data and communicates it to/from Icepack.

PR #85, opened 2018-01-04
Remove yellowstone and fix some of the documentation.

dabail10:
[Remove this and add a short summary line]:
Developer(s): D. Bailey
Other Relevant Details:


dabail10:
Oh, I see. The problem is that the italic notation is not recognized as an actual section and hence the references to them were not working. So, you object to a subsection number of 2.7.3.2.1? If not, we could use a line of carats (^) to designate the next level.

eclare108213:
Carats would be fine too.  I think there are some other subsections that are just italicized in other parts of the docs.
e

dabail10:
The italicized sections are fine if you don't need to reference them.

Dave

dabail10:
Ok. I have done the subsections at this level all with lines of carats for consistency.

Issue #74, opened 2018-01-01
documentation of bgc namelist


eclare108213:
Move zbgc namelist options into the namelist table, or create a separate table for it.  This is in the User Guide now.  Replace p1 with 0.1, etc, and remove _dbl_kind from doc table.  Is everything in the 'BGC tuning parameters' table also in namelist?

Also, the current table in the docs looks like it was copied straight out of the code (where it sets defaults).  If that is true, then some of the read values are missing _dbl_kind in the code and should be fixed.

duvivier:
Nicole, I'm going to try and reconcile these issues. I will need you to check this over though once I have it more finalized since I'm not familiar with the BGC code, namelist, etc.

njeffery:
That's great Alice.  Thanks.

I'm working now on committing some code changes to the bgc test cases. Hopefully then I'll have time for the documentation.

Nicole

________________________________

duvivier:
Nicole, please look over the updated documentation I've created here: https://duvivier.github.io/Icepack/

In particular, I need you to look over Table 8 and make sure zbgc_nml variables are correctly entered. Also, do we still need Table 9 or have these options been included in Table 8 now? 

I am aware the units for the zbgc_nml variables in Table 8 are not all formatted properly, but I wanted to make sure the information (names, units, values) was correct before delving into this.

njeffery:
Alice:


Corrections to table 8 under zbgc_nml


1. tr_zero should be tr_zaero

2. in  dEdd_algae description, remove "should not be used when sove_zbgc or skl_bgc are true"

3. in solve_zsal description,  change "therm" to "ktherm"

4. in frazil_scav, change "frail" to "frazil ice"

5. in K_Nit_diatoms, _sp, and _phaeo, and for diatoms, small plankton and phaeocystis, respectively

6. do the same for silicate half saturation,  and iron half saturation

7. from f_don_Am_protein to k_bac_l,  move the word "saccharids" or "lipids" to the end of the description.  For example " saccharids fraction of mortality to DOC" should be "fraction of mortality to DOC saccharids"

8. for lines  algaltype_diatoms to zaerotype_dust4 add to the descriptions of each after "mobile":  i.e.

 algaltype_diatoms would be "mobility type between stationary <-> mobile for diatoms".

[...]

Issue #67, opened 2018-01-01
figure and table labels in pdf


eclare108213:
Clean up labels for figures and tables, and make sure they are numbered in order.  This appears to be an issue only in the pdf version, not in the html.

Issue #63, opened 2018-01-01
Documentation chapters


eclare108213:
Fix overall layout of the Icepack documentation:
Table of contents should not be a chapter (but could be chapter 1 if necessary).
Chapter 1 Introduction (or chapter 2)
Chapter 2 Science Guide 
Chapter 3 User Guide 
Chapter 4 Developer Guide
Index of primary variables and parameters (could be chapter 6 if necessary)
Bibliography

duvivier:
This looks like an issue with the PDF generation only, the HTML looks okay to me. Does this agree with what you're seeing Elizabeth?

eclare108213:
No, I think the html shows the same problem.  "Table of Contents:" is the main heading and all of the rest of the doc is in subsections underneath that one.  I think that is coming from the index.rst file, where "Table of Contents:" seems to have a very prominent position.

duvivier:
Elizabeth,
I think I finally got this layout sorted out.
Please check the HTML:
https://duvivier.github.io/Icepack/index.html
And a PDF Generated from the same RST code:
https://github.com/duvivier/Icepack/blob/master/doc/Icepack.pdf
(Note: I will work on the labels of tables and figures next, Issue #67)

eclare108213:
That looks great!

Is there a way to control how much of the outline is shown in the left-hand panel of the html?  The individual html pages are now the subsections rather than whole chapters, and it looks like that’s also limiting the outline.  No worries if this is how it needs to be.   I’d rather time be spent on the contents, at this point, than on the layout, now that you’ve gotten the basic problem fixed.  Thank you!

duvivier:
I noticed that the left-hand panel outline had changed, but I can't figure out at the moment how to make it more extensive. I think we should wait to deal with this until the contents are finalized.

duvivier:
Being addressed with PR #90 and #91

PR #90, opened 2018-01-09
Cleaning documentation

duvivier:
Cleaning up the documentation to respond to Elizabeth's issue #63, Issue #67 , and Issue #74 . Please also check the pdf generated from the new RST under /doc/Icepack.pdf as well as the HTML. Updated year to 2018 and Icepack release to V1.0.

Removed the "sphinx" guides since that info is now in the wiki", and changed Readme.ME overall so that it points to the correct paths.

Developer(s): Alice DuVivier
Other Relevant Details:


duvivier:
Elizabeth,
I don't know. It does 194 files are changed but then when I open them it says "No Change" for most of them, including code. The ones that do show changes are the documentation files I did change. I'm not savvy enough with Git yet to know how to resolve this on my own. Thoughts? I did a git pull on the latest changes from the consortium branch immediately before I did my own commit and the PR. That's the only time I think these files would have been part of the git process.

eclare108213:
I think @apcraig might need to help on this one.  My inclination would be to start over with a fresh branch, update your fork if it needs it, then copy the changed files into the branch and add/delete as needed, commit and PR again.  But if the problem is in your fork, I'll not be much help.  There is a conflict, but probably that file was changed before being deleted when your refactored all the files.

duvivier:
My inclination is also to start over, but it seems silly to have to do this repeatedly. Maybe this is the "rebasing" thing you can do in github but I don't quite understand that fully, to be honest. I can start fresh though, that should make things cleaner for sure. I'll try to get it all done later tonight.

apcraig:
It looks like it's permissions changes on files that is being committed.  Might want to fix that.

duvivier:
I will go ahead and re-start my fork and branch to include these changes without the permissions issue. I'm going to close this PR and start again.

PR #92, opened 2018-01-10
Doc

duvivier:
Cleaning up the documentation to respond to Elizabeth's issue #63, Issue #67 , and Issue #74 . Please also check the pdf generated from the new RST under /doc/Icepack.pdf as well as the HTML. Updated year to 2018 and Icepack release to V1.0. Removed the "sphinx" guides since that info is now in the wiki", and changed Readme.ME overall so that it points to the correct paths.

Developer(s): Alice DuVivier
Other Relevant Details:

I think I got all the permissions changed so the only files changed seem to be correct this time around. The HTML for PR #91 is still good to go so that PR can be merged as well for the gh-pages branch.


duvivier:
I think it is correct that 75 files are changed. By breaking the rst documentation up we have increased the number of files. I also deleted the figure files that are no longer used in Icepack documentation (they are used in CICE documentation instead I believe), so the removal of those are a large number of the affected files.

PR #94, opened 2018-01-11
Test reporting capability to wiki

apcraig:
This adds ability to report test results to the github wiki
Developer(s): tcraig
If not, does the documentation need to be updated separately? (Y/N) Y
Other Relevant Details:

This uses a new scripts, report_results.csh to automatically push test results to the github wiki.  They are organized by hash (version) and are accumulative.  If results from a hash already exist, new results are added at the top of the results page.  If results are for a new hash, a new results page is created and it is added to the top level test results page.  This leverages the current results.csh and results.log output and parses it into an md file.  There are a few new pieces of information in the results.log file.  The wiki page is automatically checked out "clean" each time, files are added/updated, and then the changes are pushed back.

There are a few things to do still.  If adopted, we would want to have -report trigger running the script and we would probably want to remove the ctest stuff.  We need to add time zone to the date information.  I need to clean up a debug statement.   We might want to improve the top level page organization at some point, once we have a better feel how we want it organized.

One other thing, in order for the push to NOT require a user password, the following needs to be added to ~/.gitconfigure on every machine where report_results will be run in automated mode.

[credential "https://github.com/CICE-Consortium/Icepack.wiki.git"]
	helper = store

The first time the user pushes to the wiki, a username and password will be requested.  Thereafter, it will not be needed.  To see what this looks like, see

https://github.com/CICE-Consortium/Icepack/wiki/Test-Results


apcraig:
I agree the table is not ideal in github right now.  You should see a scroll bar at the bottom of each table so you can "see" the whole table if you use that.  It would be nice if making the browser wider would work, but it doesn't.  Maybe we can reduce the font size or shorten the case names or something to make the tables more readable.  I agree that would be good.  We have some control but markdown is also somewhat limiting in what we can do.  We could test embedded html.  That is supposed to work but google search suggests it's not entirely robust either.

rallard77:
Tony, This looks sharp!  Would it be possible to post the results perhaps calendar style, where you could click the date to review the results? Would it be listed by platform?

apcraig:
There are many ways we could organize the results.  I think we just have to discuss what we want to do and then see if we can get the script to do it.  I don't think it would be easy to overlay a physical calendar with the results.  That's probably asking too much.  But we could do some other things.

mattdturner:
I think this looks good.  It definitely gives us more freedom to organize / present the results in a manner that we want.  My only comment for now (which would be included in the discussion of how to organize the results) is that might be a good idea to have an `Overall` field that shows the sum of PASS and FAIL results.  This could either be on the main page (https://github.com/CICE-Consortium/Icepack/wiki/Test-Results) or above each individual table.  If the `Overall` field is on the main page, then we wouldn't even have to view the table if the sum of FAIL is 0

eclare108213:
Would it make sense (or is it possible) to group results according to when there is a bit change in the output?  E.g. there could be a number associated with the current version, say v1.0.3, which stays that way until there’s a code modification that changes the output and that becomes v1.0.4?  Big new changes, like adding a new parameterization (and tests to go with it) might become v1.1.0…

apcraig:
We still don't have a good way to label and/or tag versions.  And it's even more difficult to automate that information if it's part of our git process.  We could add a new command line option with -ts, say -tag or -label or something, that a user could (must?) include and that could be a main trigger with regard to organizing the test results.  We'd have to clarify exactly how we want the test results to be posted, how we want to name things, and then figure out how to implement it.

eclare108213:
I really like this.
Regarding the version number in columnphysics/version.txt, the documentation also has a code version number somewhere in the rst files.  Rather than having to keep them both synced with each other, maybe we should just have one?  Or do we expect the doc version number to lag the code?  It's probably easier for the reporting scripts to get the number from the doc than vice versa.  On the other hand, it's a good idea to have the version number in a directory with the code itself, in case users only carry part of the repo in their code.  @dabail10 @duvivier @proteanplanet @apcraig  what are CESM and RASM likely to do -- keep the whole repo or just the code directories?

eclare108213:
Another thought:  We have at least 2 kinds of test results to report.  First is regular, nightly (or somewhat frequent) regression testing on the code in master.  Another kind is testing of code that is sitting in someone's fork/branch, which needs to be done in support of a PR with changes to the code.  There will be a unique hash, of course, and I see that the fork and branch name are at the top of the individual reports.  I'm wondering if it's worthwhile to group the 'regular' tests separately from the development tests, on the home page.

eclare108213:
How would the quality control tests fit in the test reporting wiki pages?

eclare108213:
> we should (somehow) run a test suite on all platforms after each PR and then to also test the master each week if no PR has been executed for a week. 

I totally agree.  In generally, I don't think we would need or want to post results prior to a PR being submitted.  It should be up to the person doing the work to ensure that the basic test suite works for their computer, compiler, etc, then the PR can be submitted.  In our original vision for the Consortium, there was a minimum set of tests that the new code must pass; the results from this could just be stated in the PR.  Then the full consortium testing is done on the fork and reported.  Maybe in order to report a test done on a fork, the PR number must be identified in the submission?

Question:  can anyone from the community submit a PR (or an issue), or is that restricted to Consortium github members?

apcraig:
Anyone can push to the test results wiki.  That's how the permissions are setup now.  We can limit it to a group that we define if we want, that would not be difficult to do.

apcraig:
This has been updated, see https://github.com/CICE-Consortium/CICE/issues/71 and https://github.com/CICE-Consortium/Test-Results/wiki.

eclare108213:
I like this.  I think we should go with it, merge the PR and then refine as needed.  Any other comments or concerns?   I haven't tried to post to it yet, maybe Thurs/Fri.

duvivier:
I also like this and think we can merge the PR. My one comment would be that we should also link to the legend at the very top of the Test Results page since the sidebar link isn't super obvious.

eclare108213:
Let's make sure that any documentation changes associated with this get done.  I'm not sure what needs to be done, but there are probably places in the rst docs and the wikis that need to be looked at.

Also, I think someone else needs to try doing this, to make sure we can post results as expected.  @mattdturner could you do this?

duvivier:
Agreed. The natural places to do it would be here in the wiki, which needs updating about the test result stuff anyway. https://github.com/CICE-Consortium/Icepack/wiki/Testing-Icepack

But also in the rst that makes the doc here: https://cice-consortium.github.io/Icepack/user_guide/ug_testing.html#test-reporting

Tony, Matt, I have not tried reporting myself. If you guys can try it and write up text I can add it to the documentation or you could try changing the rst yourself.

mattdturner:
I just tried to run the code in Tony's testreport branch.  The tests ran successfully, but when attempting to submit the results to the new Test Reports page i got the following:
```
remote: Permission to CICE-Consortium/Test-Results.wiki.git denied to mattdturner.
fatal: unable to access 'https://github.com/CICE-Consortium/Test-Results.wiki.git/': The requested URL returned error: 403
```

apcraig:
@mattdturner, so far as I can tell, permissions are supposed to be setup for read/write for everyone.  Can you try to manually checkout the wiki, make a small change, and then push back.  Lets see if that works first.  The wiki is here, https://github.com/CICE-Consortium/Test-Results/wiki, and there is box in the bottom of the right hand side with the wiki repo name.  You might also try editing the wiki manually on the wiki page.  Each page should be edit-able using the "Edit" button.  Lets see if either, neither, or both of those work as a starting point.

mattdturner:
I can edit the wiki via the website.  However, if I manually clone the wiki, make an edit, and try to push back to the wiki I get the same Permission Denied error.

apcraig:
Matt and I have tested a few things.  He can edit the wiki manually without any problem, but the permissions with clone/push are not the same.  I have added him (and several others) as collaborators and now he can push.  I am looking into ways to work around this generally.  I assume we want anyone to be able to push if that can be supported.  Otherwise, we can limit it to a large group by adding collaborators as needed.  Both have pluses and minuses.  The other option might be to create a test-results user in github and to clone/push with that, very public, username/password exposed in our scripts.  That should also work if we want to go that route.

eclare108213:
This looks good, with permissions set as in https://github.com/CICE-Consortium/CICE/issues/71.
The results are posted here:  https://github.com/CICE-Consortium/Test-Results/wiki.
@duvivier please make sure the documentation is updated appropriately, as in your comment above.

Issue #58, opened 2017-12-26
restart diagnostics


eclare108213:
The heading "min/max [field]" is printed to diagnostic output when restarts are read, but the min and max values are not calculated and printed.  Either remove the heading or add the values.

PR #97, opened 2018-01-16
Codecleanup

eclare108213:
Starting to clean up the Icepack driver.  
Developer(s):  E. Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?   BFB
Other Relevant Details:
The most significant changes are to the bgc, especially the namelist.  I changed the docs to match.
Diagnostic output is different when reading restarts (min/max values are now written, addressing issue #58 ).
There's much more to do.


njeffery:
@eclare108213 : Forcing for the  bgcISPOL option is not valid beyond Dec 31.  I'm still not sure why the thermodynamics fails on Jan 1, but it's not coupled to the bgc.  With  npt = 4728, the test case works.  The restart option will still fail in the base suite because npt is 10248. Not sure what you think is best. We  could  restrict this to 4728 when testing restart for bgcISPOL.

eclare108213:
I see.  Perhaps the best thing to do is to set npt for the bgcISPOL case to be 4728 (or shorter if the actual data is for a shorter period) and remove the bgcISPOL restart test since it doesn't make sense.  I just noticed the bgcNICE is not part of the test suite, although there is a -s option for it.  @njeffery also reminded me today that we aren't testing the skeletal layer option for the bgc, but NPS has been running with that so we should be able to get a namelist configuration for it.

Issue #96, opened 2018-01-15
2 failing BGC tests


eclare108213:
See PR #89 for details

FAIL pinto_smoke_col_1x1_bgcISPOL_debug run
FAIL pinto_restart_col_1x1_bgcISPOL run-initial

eclare108213:
I think both of these tests are failing because there are 366 days of data for 2004 (a leap year) but days_per_year=365 in namelist.  This is causing the interpolation coefficients to be wrong (negative) around Jan 1, and therefore the forcing data (especially radiation) is wrong for the first timestep of 2005.  I'm not sure whether to set days_per_year=366 or to just use 365 days of data from the forcing files.  @njeffery ?  For now, I'll set days_per_year=366.

eclare108213:
The N-ICE data file also has 366 days of data, but 2015 was not a leap year, so I'm unsure what is intended here.  Logically, it makes sense to use 366 days for 2004, since we have it, and 365 days for the N-ICE data, ignoring day 366.  I'll change the code to do that.

eclare108213:
It turns out that setting days_per_year=366 is not an option, only 360 and 365 are available.  We need to test for that and either warn or abort.  If possible, it would be nice to test the leap year capability in the code using all 366 days of data (i.e. set use_leap_years=true), for the ISPOL case, if it makes sense.  @njeffery please weigh in on what the 366 days represent in ISPOL and NICE.  PR #102 at least keeps the code from crashing on Jan 1, but restarts still fail.

eclare108213:
The latest test results indicate that the bgcISPOL restart test is only failing on conrad with the intel compiler.  We here at LANL cannot debug this.  Who can?

eclare108213:
ISPOL and NICE tests are now running.

PR #102, opened 2018-01-20
BGC bug fix

eclare108213:
Bug fix for bgc forcing
Developer(s):  E Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB except bgc
Other Relevant Details:
recnum was not calculated correctly for 6-hourly forcing.  I also changed the code to only use 365 days of data, even though 366 are available in the data files.  Restarts still do not work correctly, but this fixes the Picard failures on Jan 1, which were due to incorrect interpolation coefficients.  This partially addresses issue #96


njeffery:
@eclare108213 :  I tested this with the base_suite and the individual ISPOL and NICE options.  Everything passed and looks good except the ISPOL restart.  I don't see any obvious reasons why.

eclare108213:
Satisfying this PR so that @chrisknewman can work on the restart issue

Issue #80, opened 2018-01-01
clean up index of variables


eclare108213:
Remove anything not in Icepack, and make sure all of the namelist options are included (bullets).

This one is extremely tedious, and I welcome all the help I can get.  We could divide up the alphabet...  I'll take X!

dabail10:
Holy moly. Are you trying to set the gihub issues record for 2017?

eclare108213:
Just making sure you all have something to do on Jan 2!   :)
Seriously, we know we're getting close when documentation rises to the top of the priority list.

eclare108213:
I've tried to assign all of the issues to at least one person.  Those assignments don't seem to be coming through in the email notices, but you can see them here:
https://github.com/CICE-Consortium/Icepack/issues
There are a few unassigned, looking for a volunteer......

duvivier:
I think that I should take the lead on the removal part, even though it's tedious and there are a lot of variables. I'm just afraid dividing it up might result in missing bits and pieces. What thoughts do others have?

eclare108213:
I’d be delighted if you want to take it on.  Let me know if you’d like my help.  Thanks!

duvivier:
Done with PR #106

PR #106, opened 2018-01-24
Icepack Index

duvivier:
Update to the Icepack index of variables to address Issue #80.

Developer(s): Alice DuVivier
Are the code changes bit for bit, different at roundoff level, or more substantial? Bit for Bit
If not, does the documentation need to be updated separately? (Y/N) N/A
Other Relevant Details:

All, I've gone through the Icepack index and removed variables that are no longer relevant. I just grepped through the code, so it's possible I made mistakes removing some of these or need to remove more. Please go through and double check these!


eclare108213:
Thanks @dabail10 
Some of the things you've listed should be removed, and indeed I removed some of them from the code this weekend (still in my fork).  But some of them are still in there even though we might not need them for the column physics, because they are sent through subroutine calls for other purposes. I.e. even though they might not make sense in Icepack itself, they do make sense in the context of a full sea ice model, so we've kept them.  I recommend that for each of the items noted above, we look at the code to see how it's used and whether it makes sense to remove or keep, and then do the same thing for both the code and the documentation.   I recommend to work from the code base after my fork is merged (or look at my fork, for now).

duvivier:
Elizabeth, have you updated the code from your fork where you removed some of these yet? I just wanted to check before I start a new documentation branch/pr. I didn't see icepack_index.rst changed in any of the recent PR's.

eclare108213:
@duvivier My fork has been merged into the main repo now.

duvivier:
Dave, Elizabeth. I went through Dave's comments after Elizabeth's additions. I've noted the changes I've made or the place in the code the variable lives. Do you think we should remove any of the things still in the code but that Dave highlighted?

alvdr, alvdf, alidr, and alidf --> not sure what the issue was, I'll break these out to make it clearer
daidtd, dvidtd, daidtt, and dvidtt : icedrv_flux.F90
iceumask --> removed
kstrength : icepack_parameters.F90:214
nprocs --> removed
prs_sig --> removed
Pstar : icepack_parameters.F90:223
radius : icedrv_constants.F90:70
sig1 and sig2 --> removed
swvdr, swvdf, swidr, and swidf --> not sure what the issue was. I'll break these out to make it clearer.
umask --> removed
write_ic : icedrv_calendar.F90:90

eclare108213:
Let's keep daidtd etc - might be useful even if we aren't using them now.
We might want to look at kstrength when we put in the new ridging scheme.  Pstar probably goes with it.
If radius isn't used, let's take it out of the code and the doc.
Let's remove write_ic also; it's easy enough to add another write statement manually, if needed.
The rest looks good, thanks.

PR #107, opened 2018-01-25
Add additional compilers

mattdturner:
Add PGI, GNU, and CCE compiler options for Conrad, Gordon, and Onyx
Developer(s):  Matt Turner
Other Relevant Details:
- Currently the GNU compiler fails w/ the debug option.  This is a result of non-conformance in icepack_init_zbgc.  CCE also warns of the non-conformance, but Intel and PGI proceed without even a warning.
- The results are not bit-for-bit, but we wouldn't expect bfb when using different compilers.
- In icedrv_constants.F90, nu_diag_out had to be changed from 100 to 103 due to limitations in file unit numbers with CCE.
- PGI and CCE create *.i files when running the preprocessor.  I could not figure out a way to have it directly make a *.f90 file.  As such the Makefile now has a conditional to move *.i files to *.f90 files if it is the cray or pgi compiler.


eclare108213:
Thanks @mattdturner this is a great help.  For consistency, should we also change the names of the existing files to include the (intel) compiler?

apcraig:
Thanks Matt.  There should be a way to not generate .i files.  I'll see if I can look into it in terms of what CESM is doing.

@eclare108213, I have been thinking about the machine/compiler naming convention.  I think we either have to rename all machines to include the compiler, and if we do, then I think it should be conrad_pgi or something like that.  Or, my preference is to have one file name conrad with the compiler that we prefer and then to have other machine options, conradpgi, that test other compilers but might not generally be used in production.  I can see pluses and minuses to both.  If we do move to machine+compiler convention, we could also think about trying to do that with a separate compiler flag and then having the compiler be a machine setting, so we don't end up with so many machine files.  That's probably the best long term solution.  In fact, maybe we should move to that sooner than later.

apcraig:
@mattdturner, would you be willing to update your branch with the just-merged test scripts and then run the base suites again with -report, with all the new compilers.  It doesn't have to be all machines and all compilers.  Just covering each new compiler.  Just a nice chance to share results and again check the test scripts.

Also, if we decide to split machine and compiler, that will impact our test reporting and we'll have to update it.  If we stay with machine_compiler as separate "machines", then our test results will work but the test reporting might still need some updates.  In either case, we may need to update the test reporting scripts and wiki.  It will help to see what the current scripts are doing.

eclare108213:
I think the bgc bug is a typo.  In drivers/icedrv_init_column.F90,

         f_don_in=f_don, kn_bac_in=k_bac, ...

should be

         f_don_in=f_don, kn_bac_in=kn_bac, ...

mattdturner:
@apcraig sure.  I'll go ahead and run with -report for all compilers on Onyx and Conrad

mattdturner:
I have merged the test reporting, ran the tests, and reported the results to the wiki.  The results are can be seen [here](https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_bran_forks).  I also added a commit that fixed the bgc typo, and reran the Onyx GNU case.  Unfortunately, the debug tests still fail when using GNU, with the following traceback:
```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x2020c26f in ???
        at /home/abuild/rpmbuild/BUILD/glibc-2.19/nptl/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
#1  0x20264dc3 in __ieee754_exp_avx
        at ../sysdeps/ieee754/dbl-64/e_exp.c:122
#2  0x20242bd2 in __exp
        at ../sysdeps/ieee754/dbl-64/w_exp.c:26
#3  0x20122468 in solution_dedd
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icepack_shortwave.f90:3323
#4  0x20141a44 in compute_dedd
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icepack_shortwave.f90:2797
#5  0x201479e7 in shortwave_dedd
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icepack_shortwave.f90:1362
#6  0x201520d1 in __icepack_shortwave_MOD_run_dedd
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icepack_shortwave.f90:1093
#7  0x2011a81f in __icepack_shortwave_MOD_icepack_step_radiation
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icepack_shortwave.f90:4130
#8  0x2006bfec in __icedrv_step_MOD_step_radiation
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icedrv_step.f90:761
#9  0x200055bd in __icedrv_runmod_MOD_ice_step
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icedrv_RunMod.f90:181
#10  0x2000573b in __icedrv_runmod_MOD_icedrv_run
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icedrv_RunMod.f90:56
#11  0x20002265 in icedrv
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icedrv_MAIN.f90:48
#12  0x2000235c in main
        at /p/work/turner/ICEPACK_RUNS/onyxgnu_smoke_col_1x1_debug_run1year.t00/compile/icedrv_MAIN.f90:28
```

The line in question (from the traceback) is `rdif_a(k) = (ue**2-c1)*(c1/extins - extins)/ne`.  I haven't done any debugging to see what exactly is happening.

eclare108213:
I suspect that this error is the underflow from the exponential in the dEdd shortwave, which is on the list to fix.   This is exactly why we need to run with other compilers.  Thank you @mattdturner !
@apcraig what's your thinking on the machine/compiler naming convention?

apcraig:
My thinking on the compiler settings is that we should implement a long term strategy now that would probably impact the test reporting.  Maybe we can push this change now and I can do some prototyping on that early next week or even today.  I'd like to try a few things and see what works best.  I think this task, as implemented, accomplishes the goals of porting to other machines, which is awesome.  Thanks @mattdturner for taking care of it so quick and for taking care of several compilers at once.  The errors that are generated and reporting aspects can be treated as next steps.

mattdturner:
@apcraig I just created a new issue https://github.com/CICE-Consortium/Icepack/issues/108

PR #111, opened 2018-01-30
Codecleanup3

eclare108213:
Combine icepack_constants with icepack_parameters
Developer(s): E Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB
Is the documentation being updated with this PR? (Y/N)  not yet
If not, does the documentation need to be updated separately? (Y/N)  Y
Other Relevant Details:
This is not ready yet, just creating a PR so you know it's underway.  Looks like I'll need to rebase after the previous PR is merged?


eclare108213:
@duvivier One of the changes with this PR was complete removal of the module columnphysics/icepack_constants.F90.  Reference to it also needs to be removed from the documentation, if it appears anywhere.  Its contents were moved into columnphysics/icepack_parameters.F90.  Thx!

PR #116, opened 2018-02-03
Codecleanup5

eclare108213:
Moving bgc namelist defaults into columnphysics, cleaning up.
Developer(s):   E Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?   BFB
Other Relevant Details:
Removed grid_o_t completely because it is not used anywhere (including in CICE).
The 'type' variables are needed only for setting parameters at initialization, not during the columnphysics calculations.
See https://github.com/CICE-Consortium/Test-Results/wiki/30436c16a7.pinto.intel.180203214537

### DIFF
--- a/columnphysics/icepack_algae.F90
+++ b/columnphysics/icepack_algae.F90
@@ -957,7 +957,8 @@
          Ng_to_mmol =0.0140067_dbl_kind , & ! (g/mmol) Nitrogen
          f_s = c1 , &  ! fracton of sites available for saturation
          f_a = c1 , &  ! fraction of collector available for attachment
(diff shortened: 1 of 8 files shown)


eclare108213:
Thanks.  I think this might be done.  I can still clean up other modules, later.  The thing that I don't like about this zbgc code the way it is at the moment is that many of the variables that are used in the column physics are not given default values.  These are all the arrays 'ratio_x2y' and 'zzztype' etc whose array elements are read in as namelist parameters (which are given default values in the driver, not in the column physics).  We are relying on the driver to fill those arrays.   That seems risky to me.

PR #121, opened 2018-02-05
Update comand line arguments and documentation

apcraig:
Update command line arguments to new standard, update documentation, update report_results, and add "default" options for --bgen and --bcmp

Developer(s): tcraig
Other Relevant Details:


apcraig:
Still want to update the name of "icepack.create.case", so it's not quite ready to pull but wanted to make it visible to others for further review and feedback.

apcraig:
This has been tested and is ready to merge.  Can someone do a review, then I will pull.  See

88c0faab36a6ef413d at https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks

apcraig:
Individual tests are not able to use --report right now.  We could change that.  One way around this is to create a custom test list, with 1 test in it and run it via the --suite.

I played around a bunch with the two numbers and how to present them.  The first number is the number of regression tests that failed (had different answers).  The second number is the number of tests that were not compared, either because there was no baseline or because the test failed.  
Tests generally pass or fail, but regression comparisons can pass, fail, or be undetermined.  I figured I'd add an extra integer to note how many tests were not able to be compared.  If it's greater than 1, the "green" automatically turns yellow in that space.  I played around with different headings and didn't like any of them.  My hope is that once we start seeing and understanding these tables, we don't have to have a bunch of detailed information around them.  I can add something to the legend.  Or do you think we need more?

eclare108213:
I don't think individual tests need to report, I was just wondering.  Maybe say that somewhere in the docs.
For the tables, I think you just need to add this information to the legend page.  Explain what the columns represent.  E.g. does "test fail" include both smoke and restart tests?  Or are restarts included in the "comp fail" column?

apcraig:
I have updated the legend and notes on the Test-Results page.  The documentation also is updated and is relatively clear that --report only works with --suite.  We can add some additional checking on the script to make sure --report is only turned on with --suite.  I will add an issue to do that for a future PR.

Issue #129, opened 2018-02-06
ice_brine exception


apcraig:
Hitting another error with bgc in ice_brine.F90,

  write(99,*) subname,' tcx3',-darcy_coeff, (-darcy_coeff/bphi_min*dt)
  call flush(99)
               hbrocn_new = hbrocn*exp(-darcy_coeff/bphi_min*dt)

The exp is blowing up with gnu and debug flags.  The last two values before blowup are

 (update_hbrine) tcx3  -1.1392520136377392E-009  -2.3407096177064297E-003
 (update_hbrine) tcx3 -0.56199991572839070       -3597.1318270922047     

The last set of values are creating problems with exp(-3597.).  I can keep adding checks for exps that are too large or too small, but if those values are not physical, then there are underlying issues that need to be fixed.  Thoughts?

njeffery:
@apcraig   Which test are you running?

njeffery:
@apcraig:  I've looked a bit closer at the code.  There is nothing physically wrong the exp(-3597).  The solution hbrocn_new = 0 is valid.  Go ahead and add an if  "darcy_coeff/bphi_min*dt" is larger than numerically feasible,  hbrocn_new = 0.

apcraig:
I have implemented an underflow fix that is commented out in PR #126 and will change answers.  That fix needs some review.  At the present time, this is still causing some cases to underflow when testing with the debug flags.  But the science should be ok with the underflow as we underflow to zero by design.

eclare108213:
@apcraig I think you should go ahead and fix this in Icepack, and then check it in CICE using the quality control test.  There is another exp() immediately below the one you fixed and commented out, which could be causing the other underflows.  I think what you're doing is correct.  
As you're doing this, please record all the steps you took to make the change in icepack and then do the regular and QC tests in cice, so the rest of us will know how to handle such a case.  Thx

apcraig:
Will do.  I haven't done the QC tests before.  Is there documentation or scripts?  thanks.

eclare108213:
I think @mattdturner put it into the testing scripts for cice.  Maybe -qc.  I ran it but didn't get as far as plotting the output.  It is documented in the sphinx (cice) testing section, I believe.

apcraig:
fixed with PR #158

Issue #126, opened 2018-02-05
icepack_algae underflow


apcraig:
icepack_algae is underflowing in the second part of this line with gnu compiler and debug+bgcISPOL.

               dmobile(k) = mobile(mm)*(initcons_mobile(k)*(exp(-dt*rtau_ret( mm))-c1) + &
                                 initcons_stationary(k)*(c1-exp(-dt*rtau_rel(mm))))

If anyone has any ideas, I'm happy to implement and test.

eclare108213:
This needs to be recoded.  Either rtau_ret or rtau_rel will always be zero, and so one of the exp() will always underflow.  It would be better to put the calculation of the exp in the conditional block above this, where rtau_ret and rtau_rel are computed, and then just use the values here.

apcraig:
I understand what you're getting at.  When those values are zero, I assume the exp should be treated as zero?  

So if rtau_ret is zero,
               dmobile(k) = mobile(mm)*( &
                                 initcons_mobile(k)*(-c1) + &
                                 initcons_stationary(k)*(c1-exp(-dt*rtau_rel(mm))))
and if rtau_rel is zero,
               dmobile(k) = mobile(mm)*( &
                                 initcons_mobile(k)*(exp(-dt*rtau_ret( mm))-c1) + &
                                 initcons_stationary(k)
??
Does that look correct?

apcraig:
OK.  There is another issue.  I have found another case where rtau_ret is 1.0 and ts = 3600.  The exp is under-flowing in that condition and then we'd be left with a "0" with the "exp" value so exp-1 or 1-exp would just be 1 or -1.  Mostly, rtau_ret and rtau_rel are about 1e-4 which matches up OK with ts=3600.  I need a little more help here about what to do.

There seem to be cases where rtau_* is zero and other cases where exp(-rtau*ts) truly is underflowing on large negative rtau*ts.  Maybe there are other cases too.  Can rtau*ts ever be negative?

eclare108213:
Since rtau_ret and rtau_rel are both frequencies in units of 1/s, I don't think they should ever be negative.  They are both initialized to 1 in icedrv_init_column.F90, but they should not stay that way unless bgc_tracer_type < 0.  The code comments indicate that this is actually the default condition, so this will always produce underflows in the exponential.  Now I am beginning to understand this better -- I think the intent is that the exponential should go to zero, and my comments above are not really the problem.

Is the logic is incomplete?
[...]

Issue #125, opened 2018-02-05
Abort with invalid cdn_atm_ratio_n


apcraig:
The code aborts in merge_fluxes for sets=debug,dyn with gnu.  In that case, atmbndy="constant" and cdn_atm_ratio_n is never set but it is used.

The calling sequence is something like 

icepack_therm_vertical.F90, icepack_step_therm1 defines cdn_atm_ratio_n locally
icepack_step_therm1 calls icepack_atm_boundary with cdn_atm_ratio_n as argument
icepack_atm_boundary sets cdn_atm_ratio_n when atmbndy is NOT set to "constant"
icepack_step_therm1 calls merge_fluxes and cdn_atm_ratio_n is passed in and used

If someone helps me figure out what to do, I can implement and test it.

eclare108213:
Please initialize Cdn_atm_ratio_n = c1 in icepack_atm_boundary.  
Cdn_atm can be removed from atm_boundary_const.
Thx

apcraig:
If Cdn_atm is remove from atm_boundary_const, does it need to be set to 1.0 by default in icepack_atm_boundary with Cdn_atm_ratio_n or will it be set outside in cases where atmbndy=constant?

Issue #103, opened 2018-01-21
zqsn inout


eclare108213:
zqsn should be intent(inout) in icepack_therm_vertical.F90, subroutine init_vertical_profile

dabail10:
Is this true in CICE5.1.2 as well? All that happens here is that zqsn is set to trcrn(:,:,nt_qsno,:,:). I guess Icepack does not have trcrn?

Dave

eclare108213:
It appears to be fine in v5.1.2.  In Icepack (and now CICE), depending on how the conditional is evaluated, zqsn might be used before it is defined in the subroutine.
e

eclare108213:
this one probably will change the answers

apcraig:
Fixed in PR #132

PR #132, opened 2018-02-06
fix more icepack debug errors

apcraig:
fix more icepack debug errors.  in particular,
  fix exp underflows and errors in icepack_algae (issue #126 ), icepack_brine (issue #129).
  remove cdn_atm from atm_boundary_const (#125)
  rename icedrv_restart_column to icedrv_restart_bgc (#100)
  fix zqsn intent in subroutine init_vertical_profile (#103)
Developer(s): tcraig
Are the code changes bit for bit, different at roundoff level, or more substantial? bfb except bgcISPOL
Other Relevant Details:

All tests pass on gordon, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks, hash# a9718648d.  All results are bit-for-bit except bgcISPOL.  

bgcISPOL are bit-for-bit in the log files for the full-ITD case, land, and slab gridpoints.

bgcISPOL results are different for the icefree case, diverging first here,
```
< avg brine thickness (m)=        0.09317749194835656
---
> avg brine thickness (m)=        0.09317747986181293

```

### DIFF
--- a/columnphysics/icepack_algae.F90
+++ b/columnphysics/icepack_algae.F90
@@ -67,6 +67,9 @@
       private 
       public :: zbio, sklbio
 
+      real (kind=dbl_kind), parameter :: & 
+         exp_argmax = c10    ! maximum argument of exponential
+
 !=======================================================================
 
       contains
@@ -903,8 +906,8 @@
          Sink_top,      & ! For cons: (+ or -) remaining bottom flux into ice(mmol/m^2/s)
          ocean_b,       & ! ocean_bio
          sum_react,     &
-         rtau_ret,      & ! retention frequency (s^-1)
-         rtau_rel     , & ! release frequency   (s^-1)
+         exp_ret,       & ! exp dt/retention frequency
+         exp_rel,       & ! exp dt/release frequency
          atm_add_cons , & ! zbgc_snow+zbgc_atm (mmol/m^3*m)
          dust_Fe      , & ! contribution of dust surface flux to dFe (umol/m*3*m)
          source       , & ! mmol/m^2 surface input from snow/atmosphere
@@ -933,7 +936,8 @@
                           ! when hin > hbri:  just used in sw calculation
 
       real (kind=dbl_kind):: &
-         bio_tmp          ! temporary variable
+         bio_tmp, &       ! temporary variable
+         exp_min          ! temporary exp var
 
       real (kind=dbl_kind):: &
          Sat_conc   , & ! adsorbing saturation concentration  (mmols/m^3)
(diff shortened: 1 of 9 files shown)


apcraig:
The non bfb in icefree case seems to be the change in brine implementation.  In cases where values were underflowing to zero, they are now being set to e-10.  We should make sure that value is correct.  Maybe we should change this so exp(arg)-->0.

I will rebase my branch to the current master (my base is a few versions old), modify the brine implementation to set exps(arg) to zero in the computation in the brine when underflowing, and then retest.

eclare108213:
This is good, thanks Tony.  I think that changes that are nonBFB should be brought in one at a time, in their own PR.  Optimally, they would also be tested individually in CICE before being pulled into Icepack.  That's why I've been waiting to fix bugs until we ported the new Icepack interfaces.  We might want to go ahead with this one anyhow, to save time, but in general I'd like to see the full testing procedure.

apcraig:
I refactored the icepack_brine exp underflow check and it's better.  bgcISPOL now passes bit for bit for smoke runs but not restart runs (which are longer).  I think that indicates it's a small difference, but I am happy to back it out so this PR is bit-for-bit.  See https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks.  a039157025c is the latest implementation, a9718648d702 is the prior implementation.  comp failures are down but not eliminated.

apcraig:
@eclare108213, should I back out the non bfb changes (which will continue to produce underflows with gnu)?  Do we want to test in CICE first?  Do we want to go with this implementation?  Do we want to try something else?  I am happy to go with whatever you prefer.  

I am testing the update of Icepack in CICE now.  The tricky part too is that even if icepack is bit-for-bit, that change in Icepack applied in a CICE case might not be.  We have to decide, moving forward, when we should be testing Icepack changes in CICE.  Maybe it should be anytime we suspect a non bit-for-bit change, even if Icepack is bit-for-bit?

eclare108213:
@apcraig Your last question is one that we will probably answer by trial and error over time.  Eventually we might figure out what types of changes are likely to be nonBFB in CICE even if they are BFB in Icepack.  Are you already seeing an example of that?  If so, then I think we need to test any change that we suspect might be nonBFB in CICE too, as part of the Icepack testing process.  E.g. all the bugs that I didn't want to fix until now.  I do think that in general we need to merge each suspicious one individually, so that we can more easily identify and back them out if problems come up later.  For now, I recommend that you put in as many of the BFB-in-Icepack changes as you can, but hold off on the current nonBFB ones, do them one at a time.  I hope it's not all of them but it might be.  Can I help?

apcraig:
OK, this branch is now bit-for-bit in all tests.  I have commented out the brine underflow so 1 gnu test fails, but all cases are bit-for-bit with 90ac8343f4.  See e05b6fb7fba212a3e at https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks.

PR #134, opened 2018-02-07
Suggested addition to documentation

proteanplanet:
This pull request adds a README.txt file to the doc area that I suggest should be part of the code, rather than in an unconnected web page, as to how to compile and add to the documentation system.  This is largely a replication of part of the web page.

If not, does the documentation need to be updated separately? N

### DIFF
--- a/doc/README.txt
+++ b/doc/README.txt
@@ -1,0 +1,63 @@
+THE ICEPACK DOCUMENTATION SYSTEM
+================================
+
+This README file contains a brief summary of how to add to or change documentation for Icepack. Detailed information on GitHub at:
+
+https://github.com/CICE-Consortium/CICE/wiki/Working-with-Sphinx-documentation
+
+That documentation includes an expanation of Sphinx, which we use to document CICE, as well as detailed information of how to checkout a version of Icepack, make changes to the documentation, and place a pull request in GitHub to commit your changes to the Icepack repository of the CICE-Consortium.
+
+This README file details only the information you need to compile the documentation in your local version of Icepack.  
+
+Steps for Modifying Documentation
+=================================
+
+Installing Sphinx:
+------------------
+
+This must be done once on each platform. See Sphinx or Installing Sphinx for details. On a mac laptop, execute the following on the command line:
+
+$ sudo pip install --ignore-installed sphinx
+$ sudo pip install --ignore-installed sphinxcontrib-bibtex
+
+Other platforms may require other steps. The CICE Consortium uses the following software to get successful Sphinx HTML builds including references:
+
+-python 2.7.11
+
+-Sphinx (1.6.3)
+
+-sphinx-rtd-theme (0.1.9)
+
+-sphinxcontrib-bibtex (0.3.5)
+
+-sphinxcontrib-websupport (1.0.1)
+
+You will need to use the CICE Consortium's conf.py file, which is found under /doc/source/conf.py in the repository.
+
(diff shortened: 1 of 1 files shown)


duvivier:
All, In a conversation 1/4/18 (which Andrew was not a part of) we decided I should but the "Quickstart guide for Sphinx" on the github wiki. Do we want a Readme file explaining how to use sphinx as well? The main reasoning for putting things on the wiki was to avoid having to change the info in several places if we wanted to do updates. The Readme file could just point to the wiki where we keep info on sphinx updated. 

Thoughts?

proteanplanet:
Hi Alice,
Tony and I discussed this yesterday, and we feel that the documentation should follow the code around as much as is feasible. While you've done a great job on the wiki, I have rarely encountered a code release that does not have a README.txt file to explain how to compile what is within the directory.  You'll note from the README file that I've just reduced it to the bare minimum amounting to:  "This is how to compile and produce the documentation". 
- Andrew

proteanplanet:
How about we split the difference.  In Icepack/README.md, add an entry under "## More information" with:

"-[Documenation guide] (https://github.com/CICE-Consortium/CICE/wiki/Working-with-Sphinx-documentation) Describes how to change, compile and read the documentation files in Icepack/doc"

The reason I suggested the README.txt file was because nothing is mentioned about documentation in README.md.

eclare108213:
Actually there is a link to the documentation in the main README.md, but it's not very prominent.  I've been thinking we need to set it off on a separate line, or do something so that it stands out more.  I'm not fond of the links that go to pages that just have another link on them (we have these because the sphinx info wiki is in the cice repo, so Icepack has to point to it).  I would prefer that anything that looks like technical information be in the sphinx doc itself, because if someone is already looking there for code info, that's where they might look first for similar kinds of documentation info.  So that's the direction I'm leaning at the moment.  But we certainly should add a line like the one you suggest to the README.md files, pointing to wherever the info is in the end.

duvivier:
It looks like the README.md file has a "More Information" section that has all the documentation stuff as well as the wiki. Maybe we should split that into a "Documentation" section and a "More Info" section that will have the wiki, test results, etc. We can link to the Documentation guide, as Andrew suggests, in the "Documentation" section.

proteanplanet:
I'm causing more trouble on this than its worth.  It was just not intuitive to me, based on my usual experience installing software, to not see any explanatory text docs in the install directory.  One other compromise would be to place as a comment in the top of the doc Makefile the versions of the software with which it was compiled and how to use the Makefile - basically everything from line 12 in the README.txt file down.  That's good software practice, and kills two birds with one stone.

eclare108213:
If it's not intuitive to you, then we need to do a better job of describing where to find stuff.
I do think we need a list of software, but let's not bury it in the Makefile.  Let's have it in the regular documentation.  So the question in my mind is whether it appears in the sphinx or in the wiki.  I'm still leaning toward sphinx.

proteanplanet:
I meant to add - basically I'm happy with whatever you decide is best.

duvivier:
I agree, if it is difficult for you to follow then we need to figure out how to make it more findable and understandable.

I would also lean toward adding it to the sphinx instead of the wiki. I can work on that tomorrow.

proteanplanet:
In that case, let me tell you exactly how I got into a muddle: My first reaction was to look for a README.txt in Icepack/doc.  When one wasn't there, my next reaction was just to type "make".  However, that crashed, because I wasn't using the right version of Python and Sphinx wasn't installed on the particular machine I was using.  If you can get the Makefile to throw a message pointing to the html link, or a set of instructions, when it dies, that would have fixed the problem then and there.  I note that once everything is installed correctly, the Makefile does push out the help file when one types "make".  So maybe that's another avenue here?

duvivier:
Based on our discussion, we decided to make this information a part of the developer's guide and I will add the information to the sphinx documentation and do a separate PR.

Issue #65, opened 2018-01-01
doc wind stress option


eclare108213:
Replace "We have ignored ice motion in u*..." with information about the new option that includes the ice velocity in the wind stress calculation (all turbulent fluxes?), and the circumstances under which it should be used.  This is in the 'Atmosphere' section of the Science Guide.

proteanplanet:
I can take care of this so that it caters to the high-frequency option as well.

duvivier:
Andrew, if you have text for this I can go ahead and include it in a documentation clean up I'm working on. Just let me know.

proteanplanet:
This item is complete to my satisfaction in the documentation and is awaiting as a pull request for Icepack.

duvivier:
Completed with PR #135

PR #135, opened 2018-02-07
Update to atmospheric documentation

proteanplanet:
This pull request brings the description of the physics in icepack_atmo.F90 up-to-date in doc/source/science_guide/sg_boundary_forcing.rst, and includes the addition of one reference and a change to a single comment line in  icepack_atmo.F90.  There is no change to any calculations.

Are the code changes bit-for-bit, different at roundoff level, or more substantial? BIT-FOR-BIT
Is the documentation being updated with this PR? YES


duvivier:
This looks good to me. The sphinx builds the way it should and the equations look good. Takes care of Issues #65 .

PR #139, opened 2018-02-07
Doc compilation

duvivier:
A bunch of small documentation updates (see below for details)

Developer(s): Alice DuVivier
Are the code changes bit for bit, different at roundoff level, or more substantial? bit for bit
Is the documentation being updated with this PR? Y
If not, does the documentation need to be updated separately? N/A
"Documentation" includes information on the wiki and .rst files in doc/source/, 
which are used to create the online technical docs at https://cice-consortium.github.io/Icepack/ 
Please suggest code reviewers in the column at right. 
Other Relevant Details:
Takes care of the following outstanding tasks:
Issue #73 : BGC forcing description in doc/source/user_guide/ug_case_settings.rst and  doc/source/user_guide/ug_running.rst . Will require a bit more info from @njeffery 
Issue #79 : Forcing described in doc/source/user_guide/ug_running.rst . Will require update to SHEBA from @dabail10 
PR #106 : Addressed removal of items @dabail10 and @eclare108213 mentioned in comments. See doc/source/icepack_index.rst and configuration/driver/icedrv_constants.F90
Issue #110 and PR #111 : Removed references to icepack_constants.F90 doc/source/developer_guide/dg_col_phys.rst
PR #115 : Added that updates will be part of release documentation  doc/source/intro/major_updates.rst

PR #140, opened 2018-02-07
Updated HTML for PR #139

duvivier:
HTML for PR #139 
Developer(s): Alice DuVivier
If not, does the documentation need to be updated separately? N/A
Other Relevant Details:

HTML for gh-pages branch. No code changes.


duvivier:
Tony, Yes, I think this is correct. We haven't updated the html for a while, even with other PRs. I think since this is the gh-pages branch that I'm trying to merge to there isn't really something to rebase with. Everytime we make new HTML we delete everything in this branch and replace it with what was just built in doc/build/html/  . These files are *not* included in the main repository though.

Issue #138, opened 2018-02-07
bug in icedrv_init_column.F90


dabail10:

eclare108213:
Yes.  This is why we need lots of eyes looking at the code.

Issue #136, opened 2018-02-07
Bug in icepack_atmo.F90


dabail10:
In the following code in icepack_atm_boundary:

      worku = c0
      workv = c0
      workr = c0
      if (present(uvel)) then
         worku = uvel
      endif
      if (present(uvel)) then
         worku = uvel
      endif

The second if block should be:

      if (present(vvel)) then
         workv = vvel
      endif

I have a fix coming with the next commit.

eclare108213:
good catch!  
@proteanplanet this one might affect your RASM runs?

proteanplanet:
Might just be...

proteanplanet:
Testing in RASM is underway...

PR #142, opened 2018-02-08
Unused and Dummy variables

dabail10:
This is a big one as I have removed all of the dummy and unused variables. I have attempted to just comment out cases that look like place holders. Note there are two answer changing bugs here. One is in icepack_atmo.F90 and icedrv_init_column.F90. These are issues #136 and #138.
Developer(s): D. Bailey
Are the code changes bit for bit, different at roundoff level, or more substantial? The changes are more than roundoff, but should not change the climate.
Other Relevant Details:


dabail10:
I added your suggested changes and went ahead and got rid of the SVN lines. I agree that CICE needs CCSMCOUPLED changes to CESMCOUPLED as well.

Dave

dabail10:
Don't approve yet, I have more to come.

apcraig:
Dave, is this ready to merge once we get reviews done?  Or do you have more?

PR #158, opened 2018-02-11
fix underflow in brine

apcraig:
Proposed fix to underflow in brine. 
Developer(s): tcraig
Are the code changes bit for bit, different at roundoff level, or more substantial? bit-for-bit except bgcISPOL
Other Relevant Details:

This fixes the underflow trap associated with bgcISPOL.  It also changes results by something bigger than roundoff but not science changing.


apcraig:
We are waiting for the qc test results on this.  Do not pull yet.

apcraig:
I am seeing bit-for-bit in CICE testing with bgcISPOL which is odd.

Issue #151, opened 2018-02-10
bgcISPOL dates


eclare108213:
The time of the timestep and the forcing time are not correct in the bgcISPOL runs.  It initializes to the correct date but the dates then revert to Jan 1.  The forcing seems to be for the right time, though, because the fall freeze-up happens in the spring.  I think this is just an initialization problem.

PR #160, opened 2018-02-11
ISPOL data interpolation bug

eclare108213:
Add time offset to interpolation routine
Developer(s):  E. Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB except ISPOL
Other Relevant Details:

This addresses Issue #151 and if the gods are with me, #96 

The first data record in the ISPOL and NICE data is the beginning of the field campaigns (June 17 and April 24, respectively).  The code assumed that the first record was Jan 1.  This sometimes caused the thermo to fail with bad data values.  The code now runs and restarts properly (I think).  This changes the answers for the ISPOL case.  NICE has not been tested.  

A potential lingering bug is that the ISPOL and NICE ocean data might be offset too, but variations in these fields do not matter so much.

I also commented out the snow-ice field from timeseries.csh.  It was sometimes causing the script to get into an infinite loop.  (why?)

### DIFF
--- a/configuration/driver/icedrv_forcing.F90
+++ b/configuration/driver/icedrv_forcing.F90
@@ -200,11 +200,17 @@
       real (kind=dbl_kind) :: &
          c1intp, c2intp   ! interpolation coefficients
 
-      integer (kind=int_kind) :: &  !cn ispol stuff
-          recnum, dataloc, maxrec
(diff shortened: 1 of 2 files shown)


apcraig:
Waiting for travis to finish, then will merge.

PR #161, opened 2018-02-12
Icepack doc odds and ends

duvivier:
Finalizing documentation for the release
Developer(s): Alice DuVivier
Other Relevant Details:

- Updated conf.py file for version 1.0.0
- Clarified ocean POP forcing for NICE and ISPOL
- Added functional release info
- Updated which fields can be plotted by timeseries.


duvivier:
I just added the PDF for the documentation at this point. I'm not sure if it's done - we may want to add more or modify something (e.g. the BGC POP locations). But this way we have it ready to go if we feel it is ready.

PR #173, opened 2018-02-13
Reference full path when looping over runlogs in Travis-CI

anders-dc:
v1.0.0.d0001  This PR contains a small fix that makes sure that the runlogs are correctly found if a build fails on Travis.

*Developer(s):* Anders Damsgaard, Princeton/NOAA-GFDL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ notifications:
   email: false
 
 after_failure:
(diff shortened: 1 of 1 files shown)


eclare108213:
Likewise Anders, as with Andrew's PR, we'll test and merge this one soon.  Thanks!

PR #176, opened 2018-02-14
Update test reporting

apcraig:
Update test reporting, updated format
Developer(s): tcraig
Other Relevant Details:

There was also some small formatting changes to the Test-Results page.

This also assume a string is included in the PR header of form vm.n.p.dqqqq where m,n,p, and q are integers.  v and d are the characters v and d.  This should looks like v1.0.0.d0002 and provides a naming convention for internal versions.  This is now being picked up by the test reporting script.  This string can be added just before the PR is merged anywhere in the PR title.


eclare108213:
Whose responsibility is it to add the vm.n.p.dqqqq to the PR header?  I see the next number in the test reports, so we can increment from that, provided that another PR isn't executed in the meantime.  Your most recent tests are showing all 0s...  is that intended?  https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks

apcraig:
@eclare108213, good questions.  It will be the responsibility of the person merging the PR to update the version number.  That can be done by hitting the edit button at the top right of the PR next to the PR title.  If a new version number is not added, the scripts will just continue to use the latest one that is found.  So this only needs to be done when we want to change the version number.  It will be a little effort to track the "next" version number, but as you point out, it's in the logs.  It would be nice if we could automate this task to some degree (ie. push a button and the next version number is automatically added), maybe we'll figure out a way to do that.  I'm happy to back off this approach, modify the version number format, or anything else.  Just let me know.

The latest test results are all zeros which is fine.  If you look, I ran just two tests, not the full suite to test the modifications, just to make turnaround quicker.

eclare108213:
I almost pushed the merge button without updating the version string in the header!  Have we settled on a set of guidelines for when it should be updated?  E.g. 
https://docs.google.com/document/d/1UAY-SxqaeUdB-gGbrbAADfVhE258as-wVD6p34az_dk/edit#bookmark=id.kylavbh7pozt
In particular, is "The 4th digit would only be for tracking non-BFB changes in master, not considered a formal release" agreeable?  Or do we need to track every merge?  I guess it's possible to have a merge that initially appears to be BFB but turns out later not to be, e.g. because we didn't test every option.

apcraig:
I think the idea is to update the version when we think it's reasonable.  It could be a non bfb change.  It could be an interface change.  It could be every PR for a period of time when there is a large refactor.  In general, I don't think it should be every PR always.  If we do not update the version when we should have, it's not the worst thing.  We get it next time.  As I said, I wish there were an automated or better way to do this, but ultimately, it's going to take some manual intervention one way or other.  I think we should take this versioning as helpful, but that it should not be too rigidly defined or imposed.

eclare108213:
At a minimum, let's update the number every time there is a bit- or answer-changing modification.  Then we can update at other times when we think it makes sense.

PR #181, opened 2018-03-15
Add additional Hobart support.

dabail10:
This adds some additional Hobart support for Icepack. It runs under Intel, and currently only under NAG in DEBUG mode. I've also changed some statement functions to actual functions, which is more in line with the current Fortran standard.
Developer(s): D. Bailey
Are the code changes bit for bit, different at roundoff level, or more substantial? Should be bfb on machines other than Hobart.
Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_atmo.F90
+++ b/columnphysics/icepack_atmo.F90
@@ -154,23 +154,7 @@
       real (kind=dbl_kind), parameter :: &
          zTrf  = c2                 ! reference height for air temp (m)
 
-      ! local functions
-      real (kind=dbl_kind) :: &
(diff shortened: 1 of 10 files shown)


dabail10:
Ok. It turned out to be a compiler issue with NAG 6.1. I moved to NAG 6.2 and it runs fine.

Dave

dabail10:
Made the changes as requested.

eclare108213:
Thanks Dave.  It looks like you did this from a branch on the main CICE repo rather than your fork.  I'll merge it, but in the future please work in branches of your fork...

PR #184, opened 2018-03-23
Acknowledgments

eclare108213:
Including more people who were unintentionally left out
Developer(s):  E. Hunke
Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB
Other Relevant Details:
A similar change needs to be made for the CICE docs


duvivier:
Looks good to me.

PR #190, opened 2018-04-03
Unitialized variables.

dabail10:
This commit fixes a number of uninitialized variables that were assumed to be set to zero by the compiler. It also fixes a number of intent(out) that should be intent(inout). I've also added a few trim functions for subname debugging.

- Developer(s): D. Bailey

- Are the code changes bit for bit, different at roundoff level, or more substantial? bfb

- Is the documentation being updated with this PR? (Y/N) N
If not, does the documentation need to be updated separately at a later time? (Y/N) N

- Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_atmo.F90
+++ b/columnphysics/icepack_atmo.F90
@@ -88,7 +88,7 @@
       real (kind=dbl_kind), intent(inout) :: &
          Cdn_atm      ! neutral drag coefficient
  
-      real (kind=dbl_kind), intent(out) :: &
+      real (kind=dbl_kind), intent(inout) :: &
          Cdn_atm_ratio_n ! ratio drag coeff / neutral drag coeff
 
       real (kind=dbl_kind), &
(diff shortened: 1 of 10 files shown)


dabail10:
I still have a few things to work out here. It should be bfb, but I am seeing answer changes.

Dave

dabail10:
Ok. I found the answer change. This is actually related to a bug with the initialization of zqsn. I will start a separate issue here. All tests are bfb in the base_suite except for bgcNICE.

dabail10:
Why is the travis build failing?

anders-dc:
@dabail10: The [raw log](https://api.travis-ci.org/v3/job/363237213/log.txt) shows that STDOUT gets filled with the line:

    (ridge_ice)Repeat ridging, niter =           1

This continues until the size of the STDOUT text file reaches Travis' limit of 4 megabytes.

dabail10:
So, do I need to turn off this print statement to get a valid log? Is the log file why it is failing?

Dave

anders-dc:
Disabling the repeated output of this statement will ensure that the job isn't killed because of this issue. It doesn't guarantee it will succeed :-)

dabail10:
Looks like only the bgcNICE test failed.

eclare108213:
@dabail10 can you tell why the bgcNICE test is failing?  I thought that you'd decided that the zqsn issue was a non-issue (#191).

dabail10:
Not sure yet what is going on with the bgcNICE cases. The restart test is failing for it. It looks like it was passing for Icepack1.0.0. I'll try to track it down on Friday when I'm done with the workshop.

dabail10:
So, it is not related to zqsn. There is some other variable that I have changed the initialization of that only impacts the bgcNICE case and changes answers. It also breaks restarts in the bgcNICE case. There was a time when the bgcNICE case was not exactly restarting. Does anyone remember the fix? Maybe I undid it?

dabail10:
So, the Travis test is failing with:
PASS travisCI_gnu_restart_col_1x1_bgcNICE build
PASS travisCI_gnu_restart_col_1x1_bgcNICE run-initial
PASS travisCI_gnu_restart_col_1x1_bgcNICE run-restart
FAIL travisCI_gnu_restart_col_1x1_bgcNICE exact-restart

However, my test on cheyenne with Intel is passing:
PASS cheyenne_intel_restart_col_1x1_bgcNICE build
PASS cheyenne_intel_restart_col_1x1_bgcNICE run-initial
PASS cheyenne_intel_restart_col_1x1_bgcNICE run-restart
PASS cheyenne_intel_restart_col_1x1_bgcNICE exact-restart
PASS cheyenne_intel_restart_col_1x1_bgcNICE generate icepack1.0.9
PASS cheyenne_intel_restart_col_1x1_bgcNICE compare icepack1.0.2

However, my smoke test fails the comparison test.
PASS cheyenne_intel_smoke_col_1x1_bgcNICE_debug build
PASS cheyenne_intel_smoke_col_1x1_bgcNICE_debug run
PASS cheyenne_intel_smoke_col_1x1_bgcNICE_debug generate icepack1.0.9
FAIL cheyenne_intel_smoke_col_1x1_bgcNICE_debug compare icepack1.0.2 different-data

So, what should I do here? This is only bgcNICE. Everything else passes and is bfb with the release code.

eclare108213:
Let me try to run the smoke/comparison test with the intel compiler here.

dabail10:
I'm also running the base_suite on hobart with NAG.

apcraig:
I'm struggling a little to follow exactly what's going on.  In general, the final restart files from the end of the run are compared in both CICE and Icepack to check bfb.  There are two ways things are compared.  One is against a prior set of baselines.  That has to be created separately.  The other is against another test in the same suite.  That comparison will be done as part of the suite.  The regression test is the "compare" test.  The bfb vs another test is the "bfbcomp" test.  Let me know if there is some other questions.  If there are ways to improve the implementation, lets do it.

eclare108213:
Here's what I did:

   241  9:36    git clone -b init https://github.com/dabail10/Icepack Icepack.dabail
   255  9:38    git clone https://github.com/CICE-Consortium/Icepack Icepack.master1.0.2
   256  9:38    cd Icepack.master1.0.2/
   260  9:39    icepack.setup --suite base_suite -m pinto --testid v1.0.2 --bgen v1.0.2
   263  10:03   cd ../Icepack.dabail/
   267  10:04   icepack.setup --suite base_suite -m pinto --testid dabail --bcmp v1.0.2
   271  10:19   results.csh
   272  10:19   results.csh | grep FAIL

apcraig:
@eclare108213 that looks like the correct process for testing.

apcraig:
If restart files are not written and the test scripts don't handle that properly, we obviously need to fix that.  I have not carefully reviewed each test and each configuration that it is working correctly.  There definitely could be some problems that we haven't yet identified.

dabail10:
I wonder if this is a time thing. Could it be running out of time in some cases, but not others?

Dave

eclare108213:
The ISPOL and NICE restart files aren't written because dump_freq = 'y' and the runs aren't long enough to get to a year boundary (because data isn't available).  All of the other tests have dump_freq = 'm' and are apparently working.  So the issue seems to be twofold:  (1) bgc restarts aren't being written and (2) the scripts aren't noticing that the restart files are not available, and then sometimes work and sometimes fail.  Maybe there's more to it than that.

apcraig:
If the restart files are not there, the scripts should notice that and generate a fail.  that should never result in a pass, but maybe it does.  if it does, that needs to be fixed.  i will try to have a look at that case and debug the scripts a bit.  if it's passing on cheyenne with no restarts, that's a good test case :)

Issue #194, opened 2018-04-17
Icepack code coverage


eclare108213:
The new codecov test by @anders-dc (#193) does a great job of pointing our the holes in our test coverage.  Many of the "misses" are write statements for various failure modes, which we hope never to write (unless we want to include some failures in our testing).  By far, the biggest coverage gaps are for zsalinity and skeletal-layer bgc -- I'll put the bgc ones in a separate issue.  Here's the rundown of everything else that's missing.   Maybe some of these are covered in the full base-suite, and this test was only travis?  All of these could be covered in 4 more tests (one per kcatbound value).

namelist:

- l_mpond_fresh = T
- days_per_year=360
- dumpfreq = 'd'
- calc_Tsfc = F
- default_season = 'winter', 'summer'
- update_ocn_f = T
- restore_ocn = T AND trestore=0
- atm_data_type = 'CFS'
- highfreq = T
- kcatbound = -1, 0, 2, 3
- krdg_partic = 0
- krdg_redist = 0
- tr_pond_lvl = T AND frzpond = 'cesm'
- tr_pond_topo = T AND ktherm .ne. 2
- heat_capacity = F
- conduct = 'bubbly' AND ktherm = 1

configuration:

- nslyr > 1
- ncat = 1  (kcatbound = -1)

potential bugs:

1. icepack_atmo.F90, line 652 should be .or. instead of .and.
2. why are lines 830-835 in icedrv_forcing.F90 not used?

eclare108213:
Most of these options are included in -s alt02, alt03, alt04 in PR #196, and bug 1 is fixed.  Still missing:    
* calc_Tsfc=F (I don't think this is possible in our configurations -- this is used by UK MetOffice in their coupled system)
* kcatbound=3 (I suggest we remove this option, which I couldn't get to work well enough to publish -- it'll probably generate more questions than it's worth)   
* there's some issue with tfrz_option = linear_salt  (doesn't seem to matter whether there are quotes around it in namelist, which is suspicious)
* the icedrv_forcing.F90 question remains open

apcraig:
There is an updated code coverage report based on runs on conrad with the full test suite,

https://codecov.io/gh/CICE-Consortium/Icepack/tree/586247ab794415d85ec5ca91232bb26270351cab

Lets have a look at that and decide if new tests are needed to improve coverage.  Also, I am going to create a new PR shortly to finally push the codecov capability to the Icepack consortium master.

apcraig:
[...]

PR #196, opened 2018-04-26
Addcodecov

eclare108213:
Bug fixes and extended test combinations for better code coverage

- Developer(s):   E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?

Bug fix causes tests with kcatbound /= 1 to be different.  The BGC tests are also showing different data but I think that's the script problem we've been discussing in PR #190 

- Is the documentation being updated with this PR? (Y/N)  no
If not, does the documentation need to be updated separately at a later time? (Y/N)  yes, if we keep these tests

- Other Relevant Details:

Bug fix 1:  the Icepack driver was not reading grid_nml so was not resetting kcatbound from the default, which is 1 in the code.  I reset the default in icepack_in to be 1 so that the current tests would not all fail.  

Bug fix 2:  as noted in #194, there was a bug in icepack_atmo.F90 ('or' not 'and').  However the quantity being calculated was never used so I fixed the bug by removing those lines completely.

Added several new test combinations which capture most of the code coverage issues in #194.  These run but the output needs to be looked at carefully to see if they are doing what we expect them to do.


eclare108213:
I named the new options alt02, alt03, alt04, but we could come up with a more descriptive set of names.  E.g. alt02 is very low resolution within the column (1 layer in the vertical, 1 category); alt03 is higher snow resolution (5 layers) but otherwise much simpler (and summer conditions so maybe this doesn't really test the snow resolution).   We do need to look at all of our test combos to make sure they make sense.

apcraig:
These changes look fine.  Before approving, a couple thoughts.  First, I agree something more descriptive than alt01, alt02, etc might be better.  Second, we probably need to come up with some general rules about whether the test names must change when the test changes.  So, lets say we are running alt01 and then we decide to change the forcing or anything else about the alt01 test.  And that changes answers.  Is it OK to continue to call it alt01 or does alt01 need to be retired and we create a new test.  In some ways, we would want to do that for transparency as well as regression testing.  If alt01 shows a regression fail when we expect it to, that's not particularly useful.  It might be better to stop testing alt01 and start testing alt01a and the regression testing will then not show a fail.  Finally, it sounds like there is still some tuning to be done with these tests to get the setup "best".  Should that occur as part of this PR or as part of continue development?

apcraig:
If the code changes and answers change, then that's what we're testing for and test names should not change.  I believe even in the case where a test namelist also has to change.   I think the only situation where a testname might change is if the test changes and the answers change as a result of that.

eclare108213:
I agree with that

eclare108213:
@apcraig  Icepack could benefit from the same testing organization that you are undertaking for CICE.  In that case, we might want to abandon this PR in order to avoid the renaming issue, or we could merge it then change things, or we could take a middle route and rename the tests temporarily here, merge the PR, then name them using alt01, alt02 etc following the same pattern as in the CICE tests.  Lots of options, we just need to keep moving forward.  Preference?  And yes, these tests should be added to the base_suite (or similar).

apcraig:
I think we should keep this PR moving forward.  It's too hard to keep the CICE and Icepack scripts synchronized at this point.  If we decide we want to update the Icepack scripts with some of the features of the CICE scripts, we can do that, but lets make it a separate PR and maybe even create an issue first outlining what features are desired.  I think the scripts are going to slowly diverge at this point, and it's going to be hard to prevent that.  

I will again throw out a suggestion that we consider merging Icepack back into CICE, bringing the Icepack driver with it, making the Icepack driver a separate configuration within CICE, and then having a single set of scripts.  There is cost associated with maintaining two separate repositories, both with regard to having two scripts packages and also the git submodule stuff.  Maybe we can revisit this in the future, I think now is too soon.  I would like to get some feedback from E3SM on the Icepack separation, new interfaces, Icepack scripts, and so forth before we do anything.  If E3SM is keeping their own copy of column physics and has little interest in regularly contributing/merging/testing/etc with the tools/repo provided by the consortium, then maybe the consortium, in a year or two, can think about what works best for the consortium.

eclare108213:
Ok, let's merge this one now, then add the new tests to the base suite and check code coverage, and go from there.

apcraig:
Are you able to add these tests to the base_suite before we merge the PR?  If you need some help, let me know.  You should be able to just add one line per test to the base_suite.ts file.  You can even rearrange or remove some other tests if it makes sense with respect to coverage.

apcraig:
the automated testing is testing the travis_suite, so we're not changing that.  Also, if we do automated testing outside travis, which we aren't quite yet, adding new tests is ok.  if we do regression testing, it just will show there wasn't a baseline.

eclare108213:
I added alt02, alt03, alt04 to the base_suite, both for smoke and restart tests.  The new tests pass except that 03 and 04 fail restarts, so I've commented those out for now.  The purpose of this PR is to get the tests set up; a separate step will be debugging the new restart tests.  

I fixed the initial condition, which had a hardwired category-3 value that created problems for 1-category tests.

Issue #178, opened 2018-02-21
BGC skeletal layer test


eclare108213:
Create a test for the BGC skeletal layer configuration

PR #199, opened 2018-05-10
Corrects bgc skeletal test case and modifications for new ISPOL/NICE files

njeffery:
Adds option bgcsklNICE:
1. initializes bio fluxes and other diagnostic fields: BFB
2. corrects growth rate diagnostic: BFB

For options: bgcISPOL, bgcNICE and bgcsklNICE:
1. updates offset and data location for new ISPOL/NICE forcing files
dated 05092018: nonBFB with the above options

@eclare108213 , @duvivier :
This PR should run with the old files NICE_atm_forcing.txt and ISPOL_atm_forcing.txt, but there were some odd offsets in those files.  Results are better with NICE_atm_forcing_05092018.txt and ISPOL_atm_forcing_05092018.txt.  @duvivier, I'll send you the updated files.

Also, the doc/make.bat file has changed without my intent.  Is this expected?
The description of ISPOL/NICE forcing description doesn't need to be updated.


eclare108213:
@duvivier @apcraig @dabail10 (and anyone else):  I don’t remember that we have a plan for updating data files. Do we keep the same names for the new ones and relabel the old ones, or vice versa?  We should definitely keep a log of changes in the wiki, on or linked from the test data page.

eclare108213:
https://stackoverflow.com/questions/5834014/lf-will-be-replaced-by-crlf-in-git-what-is-that-and-is-it-important

This explains that the issue is formatting differences between Unix and Windows for the end of file line feed.  There is a git command that fixes it.  Maybe we should do that, @apcraig?

eclare108213:
This PR is waiting on responses regarding how to handle the change in bgc forcing files (how to name them, where to put them, what to do with the old ones) and also the make.bat issue (see also 
http://lexsheehan.blogspot.com/2014/05/warning-crlf-will-be-replaced-by-lf-in.html).  
@duvivier @apcraig @dabail10 please express your opinions...

eclare108213:
I agree we shouldn't overwrite files.  Can we do this using soft links on an ftp site and in the tar files?  I'd rather not have to change the code every time a file name changes.  Perhaps a better solution would be to put the filenames in the namelist -- that way the change is explicit in the code base and users can choose to continue using an older file, provided it's not fundamentally different.  What should we use as a file naming convention?  Date?  version?  hash?

apcraig:
I think the date string works fine.  Some sort of versioning is fine too.  In CESM, we have often used file_yymmdd.xyz but it's not a requirement.  The requirement is not to overwrite files.  Any name will do.  We should not have any hardwired files in the code, they should all be in namelist.

eclare108213:
To keep this moving, @duvivier would you add the new files to the ftp site?  @njeffery already has dates on them.  (maybe you already did that?)   I'll create an issue to put the filenames in namelist. 

@apcraig there's still a question about make.bat, for which I'd really like your expert opinion...

duvivier:
@eclare108213 @apcraig I just moved the files to the FTP and left them named as Nicole has them. They are in the tarred and zipped Icepack_data.tar.gz file and under /forcing/ISPOL_2004/ and /forcing/NICE_2015/ respectively. The original ISPOL_atm_forcing.txt and NICE_atm_forcing.txt files are still in the directory as well once you untar the files.

apcraig:
Regarding the make.bat, I assume @njeffery is using a windows system.  I can't test any of this, but it seems he needs to set 

git config --global core.autocrlf true

to make this problem go away.  Separately, one of us could fix this later, but it would be better if the change were never checked in.  @njeffery, are you working on a Windows system and what does

git config --list 

produce on your machine?

eclare108213:
I have tested this branch and compared with the current code.  The physics is BFB.  I expect the bgc test cases to be different with the new forcing files, and they are.  ISPOL looks okay (it's in the southern hemisphere, right?), but I don't think things are configured correctly for NICE.  E.g. here is a plot of the shortwave forcing for NICE:
ice_diag full_itd_shortwave_radiation_sum

njeffery:
@eclare108213, the NICE forcing is only valid from April 24 - June 6, which is the segment that looks ok in your figure. For an hour time step, this is istep0 = 2736 and npt = 1032.

eclare108213:
Then let's set this one up so that other dates are never used with the NICE data.  There must be data available for the whole year, otherwise how can it run?  This is very confusing.

eclare108213:
I will merge this PR and fix the dates for the NICE tests in my own fork.  I'll also add the skl tests to the test suite.

eclare108213:
addresses issue #178

PR #203, opened 2018-06-06
bgcNICE tests and bug fix

eclare108213:
bgcNICE tests and bug fix

- Developer(s):  E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?   BFB

- Is the documentation being updated with this PR? (Y/N)  YES
If not, does the documentation need to be updated separately at a later time? (Y/N)   N

- Other Relevant Details:

The NICE data is only valid for 24 days, so restart tests do not make sense.  I removed them from the base and travis test suites.

Added a skeletal layer BGC smoke test to both test suites.  It also uses the NICE data.

set_env.bgc is not used, so it's been deleted.

Fixed a bug in the skeletal layer configuration that was causing the code to crash in debug tests with a trcrn index = 0.

Checked that the timeseries plotter works correctly for NICE output.

Base_suite passes all tests including the new skeletal layer smoke test.  See https://github.com/CICE-Consortium/Test-Results/wiki/v1.0.0.d0003.7b21edfc32.pinto.intel.180605.235425


eclare108213:
Added to this PR:
Moved forcing filenames to namelist and updated the documentation for the new namelist fields, atm_data_file, ocn_data_file, ice_data_file, bgc_data_file.  The default icepack_in sets these new fields consistently with the default forcing used (clim with SHEBA ice convergence). Updated forcing filenames for ISPOL, NICE to the new ones provided by @njeffery (which are available on the ftp site).  The base suite passes as before.  Have not checked the documentation results.  
This PR should be ready to merge now, if the reviewers agree.

duvivier:
Should bgc_data_dir and bgc_data_file both be specified in the bgc namelist instead of split into the bgc and forcing namelists? 

Otherwise this looks fine.

eclare108213:
@duvivier Good point.   bgc_data_dir is not used at all in Icepack, so I'll remove it.  We might need it for CICE, though, and in that case I think it should be in the forcing namelist rather than bgc, to make it consistent with other parts of the code (e.g. we don't have the SHEBA convergence data flag in the dynamics namelist).

eclare108213:
@duvivier I removed references to bgc_data_dir from the documentation.  If that's all that's needed for the docs, then I think this one is ready to merge -

duvivier:
Okay, I think that was all that was needed. Merging now!

PR #207, opened 2018-06-24
update scripts for automated testing 

apcraig:
and other features consistent with CICE scripts

- Developer(s): tcraig

- Is the documentation being updated with this PR? (Y/N)  N
If not, does the documentation need to be updated separately at a later time? (Y/N)  Y

- Other Relevant Details:

This updates the Icepack scripts to be consistent with the current version of CICE scripts and to allow automated testing easier.  The main new features are the ability to set the version string and to handle multiple test suites with a single submission.  In addition, quiet mode was added and turned on for travis, reporting was updated, and setup_run_dirs.csh was added.  This should now be consistent with the CICE scripts implementation.

Once this is merge to the consortium, I will setup automated testing.


apcraig:
@anders-dc  , travis is failing, but I'm not sure why.  I did not change code but did change the scripts.  I'm sure I broke something but not sure what.  Any thoughts?

apcraig:
See scriptsB branch 4ed9dba5cf44ae02820d3655 in the test results for a full suite run on conrad,

https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_bran_forks

apcraig:
bgc restarts seem to be broken, should they be working?

apcraig:
I just checked whether there are new input datasets for BGC and the tar file seems to be the same with what I have downloaded on conrad.  Were any bgc datasets recently updated to get bgc working in Icepack?  The current version of Icepack on the trunk also fails the bgc tests on conrad.  So my branch is identical in results.  Just trying to figure out if I'm missing something.

apcraig:
@anders-dc, I fixed the travis problem.  After staring at the output a bit, I figure out the problem.  The test suite directory name was changed with the scripts update, so now we're passing.

I think the only outstanding question is whether bgc should be passing.  It is not in my branch and it also is not on the current trunk on testing on conrad.

apcraig:
Working on another set of changes, hold off on review and merge.

eclare108213:
I just checked that the master base_suite passes for me, on pinto:  https://github.com/CICE-Consortium/Test-Results/wiki/v1.0.0.d0003.4ed9dba5cf.pinto.intel.180625.180641.  The forcing changed, and there are new bgc files with different names, e.g. ISPOL_atm_forcing_05092018.txt.  Also, NICE restarts should not be tested.  See PR #203.  Do you have all of those changes?

eclare108213:
I've also tested this branch on pinto.  Must be something specific to conrad.
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_bran_forks

apcraig:
Thanks Elizabeth, Rick just found the same issue.  We had outdated input datasets on conrad and gordon.  @duvivier , I'm going to recommend that maybe we add a date string to the tar files on the ftp site indicating the date they were created.  That way it's easier to tell if a "relatively newer" input data tar file might be on the ftp site.  But that might also cause problems for travis, which is downloading the "current" file when it runs.  Maybe we need both a generic and a date specific tar file or a link between them.  I think as long as the tar files are cumulative (we don't delete data generally), we don't need to keep old tar files around, so this should not result in a large increase in data on ftp.

I have updated the input data on conrad and gordon and am running some tests, but suspect that should fix the problem.  In a broader sense, we don't really have a way to tell whether the local input data is up to date for the current code on any particular machine nor how to communicate with the community that input data has been updated and relating the code versions and input data at any level.  That's definitely a shortcoming.  Maybe we can think about if/how we can do that.  It's relatively low priority, but something to think about.

Maybe we can also implement a script that will quickly update the input data.  Something that can be run quickly by any users, that downloads and untars the data into the local input data area.  I'll see if I can put something like that together.  In fact, we could then add a new option to create_newcase that would run that script.  That could be quite handy.

apcraig:
Let me update the documentation then we can merge.

apcraig:
Ready for review and merge?

eclare108213:
Looks okay to me.  I'll merge.

PR #219, opened 2018-09-06
CMIP changes

dabail10:
This PR is necessary for the CMIP changes (CICE Issue 93) and also add changes for the evaporation over snow and sea ice issue (CICE issue 97). The CICE related CMIP changes are coming in a future PR. Here is a short summary of what has changed:

1. fcondbot and fcondbotn are added so they can be shared with the CICE code.

2. Tbot is now shared with the driver and hence CICE.

3. A new diagnostic variable Tsnic has been added for the snow-ice interface temperature.

4. New variables evaps and evapi along with category versions have been added to keep track of evaporation over snow and bare ice separately.

- Developer(s): D. Bailey

- Is the documentation being updated with this PR? (Y/N) N
If not, does the documentation need to be updated separately at a later time? (Y/N) N

- Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_flux.F90
+++ b/columnphysics/icepack_flux.F90
@@ -32,18 +32,22 @@
                                strairxn, strairyn,   &
                                Cdn_atm_ratio_n,      &
                                fsurfn,   fcondtopn,  &  
+                               fcondbotn,            &
                                fsensn,   flatn,      & 
                                fswabsn,  flwoutn,    &
                                evapn,                &
+                               evapsn,   evapin,     &
                                Trefn,    Qrefn,      &
                                freshn,   fsaltn,     &
                                fhocnn,   fswthrun,   &
                                strairxT, strairyT,   &  
                                Cdn_atm_ratio,        &
                                fsurf,    fcondtop,   &
+                               fcondbot,             &
                                fsens,    flat,       & 
                                fswabs,   flwout,     &
                                evap,                 & 
+                               evaps,    evapi,      &
                                Tref,     Qref,       &
                                fresh,    fsalt,      & 
                                fhocn,    fswthru,    &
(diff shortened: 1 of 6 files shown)


dabail10:
I was thinking  T sn(ow) ic(e). Would you like Tsnowice better? I would definitely like the CMIP-CICE changes to go in soon. I need the first pull request with the dummy stuff merged before I can do this though. The documentation will only need to be updated for CICE I believe, not icepack.

eclare108213:
Tsnowice sounds like we have a temperature for a snow-ice layer, which we don't resolve (yet!).  How about Tinterface?  I'd turn that into Tintfc...

dabail10:
I was concerned about the use of intfc, because if the icepack_intfc.F90 module.

dabail10:
I did not find Tsnice anywhere, so that is good.

Dave

dabail10:
Renamed Tsnic to Tsnice.

apcraig:
@eclare108213, can you approve or comment then I'll start merging.

eclare108213:
This looks fine to me.  There is a single change to the shortwave module, which is commented out.  It's better not to have stuff like that in there unless there's a meaningful reason for it.  But I'm okay with merging this now.

dabail10:
This changes answers, so I am waiting to commit it separately.

Dave

apcraig:
Just to be clear, it's that shortwave change that is commented out that changes answers, right?  That's a little scary, it's just an intent change.  Does is change science?  I guess we can talk about it when the change is proposed later.

dabail10:
Exactly. I am unsure so far why it changes answers. It is something to do with the setting the initial snow depth and fraction related to the level ponds. These are set to zero before the call to shortwave_dEdd_set_snow, so they should be intent(inout) in this subroutine.

apcraig:
I just had a quick look at the code too.  I agree they should be in/out.  But it's odd that it changes answers and that the intent issue wasn't picked up before now.  Interesting.  thanks.

dabail10:
The intent inout was only picked up by uninitialized variable checking with the NAG compiler.

PR #220, opened 2018-09-11
Add dump_last feature to icepack.

dabail10:
This adds the functionality to icepack to write a restart on the last step of a run. By default it is off. It will be turned on for smoke tests. Note, this is on top of the CMIP.

- Developer(s): D. Bailey

- Please suggest code Pull Request reviewers in the column at right.

- Are the code changes bit for bit, different at roundoff level, or more substantial? BFB

- Is the documentation being updated with this PR? (Y/N) N
If not, does the documentation need to be updated separately at a later time? (Y/N) Y
Note: "Documentation" includes information on the wiki and .rst files in doc/source/, 
which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-icepack/.  

- Other Relevant Details:

PR #221, opened 2018-09-11
Add dump_last namelist feature and turn it on for smoke tests.

apcraig:
Add dump_last namelist feature and turn it on for smoke tests by default.  Some bgc tests were not producing any restart files so no comparison or regression testing was working.  See issue #220. 

- Developer(s): tcraig

- Is the documentation being updated with this PR? (Y/N) Y

- Other Relevant Details:

Tested on conrad with intel, pgi, cray, and gnu with all bgc tests as well as the quick_suite, bit-for-bit.  Verified restarts are created and regression testing works with smoke bgc tests.

### DIFF
--- a/configuration/driver/icedrv_calendar.F90
+++ b/configuration/driver/icedrv_calendar.F90
@@ -88,6 +88,7 @@
          new_hour       , & ! new hour = .true.
          use_leap_years , & ! use leap year functionality if true
          write_ic       , & ! write initial condition now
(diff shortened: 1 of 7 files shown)


apcraig:
Should we wait for the CMIP PR to be closed before doing this one?  I don't think there will be conflicts, but am also happy to wait.  I guess we are waiting on one more review for the CMIP PR?

dabail10:
I think this one can go in anytime.

PR #227, opened 2018-09-27
Intent inout in shortwave_dEdd_set_snow

dabail10:
This fixes a bug where the variables fs and hs should be intent inout in shortwave_dEdd_set_snow. This change is bfb in icepack except for hobart_nag_restart_col_1x1_alt03. This is with ktherm=1 and the topo poinds. All of the hobart_intel tests passed and were bfb. I'm not sure why this one case is not bfb.

- Developer(s): D. Bailey

- To verify that this PR passes the initial QC tests, lease include the link to test results

- Is the documentation being updated with this PR? (Y/N)

- Other Relevant Details:

### DIFF
--- a/columnphysics/icepack_shortwave.F90
+++ b/columnphysics/icepack_shortwave.F90
@@ -3482,8 +3482,7 @@
          Tsfc   , & ! surface temperature 
          hs0        ! snow depth for transition to bare sea ice (m)
 
-!    real (kind=dbl_kind), intent(inout) :: &
-     real (kind=dbl_kind), intent(out) :: &
+     real (kind=dbl_kind), intent(inout) :: &
          fs     , & ! horizontal coverage of snow
          hs         ! snow depth
 


dabail10:
Running the CICE base_suite again.

Dave

dabail10:
So, I just compared back to some CICE baselines on cheyenne awhile ago. All of my tests are different with CICE master. So, maybe this is a compiler change on cheyenne? I guess I will go back and regenerate the baselines from CICE master again.

dabail10:
Ok. So, 27 out of 32 compares failed in the CICE base_suite with this single change. The issue is the following in pseudo-code:

fs = c0
hs = c0

call set_snow(fs,hs)

...

later use fs/hs in a calculation

subroutine set_snow(fs,hs)

intent (in) fs,hs

hs = vsno / aice

if (hs >= hs_min) then
   fs = c1
  if (hs0 > puny) fs = min(hs/hs0,c1)
endif

So, when fs and hs are intent(in), this means the initialization earlier is ignored. In the subroutine "set_snow" hs is alway set, but fs might not be, i.e. when hs < hs_min. So, imagine a situation where the first time through hs = hs_min + eps. Then fs is set according to the min statement. However, the next time through maybe hs = hs_min - eps. Then fs is not reset to anything and it remembers it from the last call.

However, when fs and hs are intent(inout), they are always reset to zero before the call to "set_snow". I believe this explains the answer changes.

PR #231, opened 2018-11-19
update icepack version in preparation for 1.1 release

apcraig:
Update the icepack version number via

./icepack.setup --setvers v1.1.0

- Developer(s): tcraig

- To verify that this PR passes the initial QC tests, lease include the link to test results

- Does this PR create or have dependencies on CICE or any other models?  No, but will be used in CICE release.

- Is the documentation being updated with this PR? (Y/N) Y, version number

- Other Relevant Details:

Non code changes, no tests were run.  Will run a full icepack and eventually test suite upon merge.


apcraig:
I don't think this PR is the final update before Icepack1.1 release, just a step.  I am happy to update some documentation if you'd like.  Just let me know what you'd like me to do.

What I want to do is finalize the release version on the trunk then create the release branch and tag, both for Icepack and CICE.  I prefer not to get into a situation where we create a release branch, update the branch a few times, then have to merge those changes back to master.  Better to stay on master until the release is ready.  We just have to pick a point where we call it ready.

duvivier:
In that single documentation file (doc/source/intro/major_updates.rst) we explicitly say that "this is release version 1.0" so we should at least change that. I don't know if we want/need to add a lot more to the major updates section or if we'd rather do that in the release notes with the new tag.

apcraig:
Good catch.  I have removed that line entirely.  My feeling is that we should avoid sentences that require maintenance :).  We could potentially have some other "seds" in the documentation that automatically change when we execute icepack/cice.setup --setvers.  Right now, we change a version number in two places, we could add more.  Let me know if we want something like that.  Happy to do it.

apcraig:
The versioning is not ideal.  What would be great would be increment the minor number anytime answers change or to incorporate the hash directory into the source code and documentation, of course, all automatically.  Neither of those are practical.  We have chicken and egg issues.  You can't document the hash until after the PR is done.  I also don't think it's practical to try to maintain manually changing versions as perfectly defined points in development.

I believe we should do the best we can.  We update version number when appropriate, knowing that we might set it a few commits before the "official version".  We also know that version number will stick until we change it again.  It might be fixed for months and many answer changes.  Again, maybe not ideal but OK.

I think our primary focus should be for the version number to be correct in the releases.  The version number on master will track code changes on some level but that they change at clearly defined gates is too complicated to deal with.  On master, we can track via hash.  Again, not ideal because the randomness is hard to deal with, but it's git.

If folks checkout random versions from master or elsewhere, they just have to know where they are in evolution of the model.  If they know the hash, we can help them.  I think we should assume most people will work off releases or off master.  And the process we have is OK for both, I think.

Happy to consider other ways to handle it and I could be missing some easy improvements.  But I also think we need to keep in mind effort vs gain.

apcraig:
Also, just to be clear about one other thing.  The release will be a tag in github.  That tag will be tied and fixed to a particular hash.  We will have a release branch, mostly for ease of maintenance.  The version name in the model and model documentation is just a convenience thing.  It doesn't really formally tie a version to a hash, the git tag does.  At least that's how I think about it.

eclare108213:
This makes sense to me.  And since the title of this PR is explicit about this being preparation for release, if someone looks at log info for the branch or master, they should recognize that it's not the release itself.

PR #235, opened 2018-11-28
doc mods for Icepack release

duvivier:
Finalizing documentation for Icepack (and CICE) releases. 

Developer(s): Alice DuVivier 

Does this PR create or have dependencies on Icepack or any other models? No

Here are the readthedocs tests of the Icepack documentation changes:
https://icepack.readthedocs.io/en/latest/index.html
https://media.readthedocs.org/pdf/icepack/latest/icepack.pdf


doc/source/user_guide/ug_troubleshooting.rst
```diff
@@ -101,13 +101,13 @@ melt pond albedo ``albpnd`` is only computed for the Delta-Eddington
 shortwave case.
```
duvivier: @dabail10 can you check if this part in particular looks right?

duvivier:
@eclare108213 

>doc/source/science_guide/sg_boundary_forcing.rst ? Is this duplicated somewhere else? It's a column issue, so should be here rather than CICE. 
I think I added the correct text back in doc/source/science_guide/sg_boundary_forcing.rst. Please check the html: https://icepack.readthedocs.io/en/latest/science_guide/sg_boundary_forcing.html
Now "calc_Tsfc" should be present twice.

I removed the colon.

PR #244, opened 2018-12-01
fix readthedocs url

eclare108213:
Testing the process for testing the documentation using readthedocs revealed some URL errors

- Developer(s):   E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB

- Is the documentation being updated with this PR? (Y/N)  yes

- Other Relevant Details:

https://icepack-eclare.readthedocs.io/en/doc1/index.html

This needs to be fixed in the CICE docs too.

The instructions are very clear up to the point where you want to look at documentation for a branch rather than master.   I managed to click around until I found it, something like this:
Click on Versions, find your branch in the "Inactive" list, click 'Edit' then check the 'Active' box and save.  Then click on 'Builds' and the branch name.   Is there an easier way?  If not, maybe we should add these instructions to the docs (both Icepack and CICE).

I might add more commits here -- don't merge yet please.

### DIFF
--- a/doc/source/developer_guide/dg_documentation.rst
+++ b/doc/source/developer_guide/dg_documentation.rst
@@ -7,7 +7,7 @@ Documentation System
 
 With Icepack development, corresponding updates or modification to the Icepack
 documentation are required. Whenever you modify the model you should update
-documentation. Icepack uses `readthedocs.org <readthedocs.org>`_ to create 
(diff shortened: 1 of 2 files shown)


eclare108213:
Icepack docs look good.  This is ready to be merged.

PR #245, opened 2018-12-03
how to add a tracer to Icepack

eclare108213:
Updated instructions for adding a tracer to Icepack column physics and driver.

- Developer(s):  E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?   BFB

- Does this PR create or have dependencies on CICE or any other models?  no

- Is the documentation being updated with this PR? (Y/N)  yes

- Other Relevant Details:

These steps should be checked by someone actually adding a tracer, to see what might be missing or confusing.  A similar set of steps is included in the CICE documentation,
https://github.com/CICE-Consortium/CICE/pull/266/commits/e8fac2d1b254aa8512b70655e59f892e66ed6cef


apcraig:
I'm not sure how far to go with testing "adding a tracer".  I think that's a fairly sophisticated process that requires adding a bunch of code various places.  And it seems like the code needs to be a little meaningful to test the process.  I guess we could test adding a new tracer that does little and then make sure it doesn't change answers.  There is an issue in CICE (238) for doing more technical testing on the tracers in CICE and other refactoring.  Maybe we should add this task to that issue?

apcraig:
I just updated CICE 247 and added a task for testing/documenting ability to add a new tracer.

eclare108213:
I wasn't proposing a distinct test for this, just that someone check the docs in the course of their work when doing some future model development with a new tracer.  I will be adding new tracers in a few months... Thanks for adding this item to CICE issue 238 -- the first extension will be a good user-friendliness test of the Icepack interface.

apcraig:
@eclare108213, I agree with regard to new tracer testing, I think we're on the same page.  I don't think we can have a "canned" test for it, just that we should try to test the ability to add a tracer manually at some point.

PR #254, opened 2019-03-08
update cheyenne modules for machine upgrade

apcraig:
Update cheyenne, required with machine upgrade early March, 2019

- Developer(s): tcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial?  bit-for-bit, see a838da5 at https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks

- Does this PR create or have dependencies on CICE or any other models?  No

- Is the documentation being updated with this PR? (Y/N) No
If not, does the documentation need to be updated separately at a later time? (Y/N) No

- Other Relevant Details:

Full test suite is bit-for-bit with last weekend test. Old modules no longer supported with machine upgrade in early March, 2019. These changes should be usable in older versions of Icepack as well.

### DIFF
--- a/configuration/scripts/machines/Macros.cheyenne_intel
+++ b/configuration/scripts/machines/Macros.cheyenne_intel
@@ -33,24 +33,25 @@ NETCDF_PATH := $(NETCDF)
 PIO_CONFIG_OPTS:= --enable-filesystem-hints=gpfs 
 
 #PNETCDF_PATH := $(PNETCDF)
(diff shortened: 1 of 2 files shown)


apcraig:
I will merge the PR later this evening even without a review.  Cheyenne is currently not working on master, these have been tested to be bfb in a full suite, and these changes only affect cheyenne.

PR #257, opened 2019-04-24
Move makdep compilation to Makefile

phil-blain:
* makdep compiled from a makdep rule in the Makefile
* Add CFLAG_HOST macro
* makdep is only compiled if needed
* This addresses CICE-Consortium/CICE#306

- Developer(s): Philippe Blain

- Please suggest code Pull Request reviewers in the column at right. @apcraig 

No tests, did not modify source code.

- Does this PR create or have dependencies on CICE or any other models? No

- Is the documentation being updated with this PR? (Y/N) Yes
[https://icepack-phil-blain.readthedocs.io/en/build-system-robustness/user_guide/ug_running.html#cross-compiling](https://icepack-phil-blain.readthedocs.io/en/build-system-robustness/user_guide/ug_running.html#cross-compiling)
If not, does the documentation need to be updated separately at a later time? (Y/N) No

- Other Relevant Details:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,17 @@ language: cpp
 
 sudo: false
 
-compiler: clang
-
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-      - llvm-toolchain-trusty-5.0
     packages:
       - tcsh
       - pkg-config
       - netcdf-bin
       - libnetcdf-dev
       - gfortran
-      - clang-5.0
+      - gcc
       - wget
       #- lftp
 
--- a/configuration/scripts/Makefile
+++ b/configuration/scripts/Makefile
@@ -114,7 +114,7 @@ ifndef $(CFLAGS_HOST)
   CFLAGS_HOST :=
 endif
 
-makdep: $(ICE_CASEDIR)/makdep.c
+$(DEPGEN): $(ICE_CASEDIR)/makdep.c
 	$(SCC) -o $@ $(CFLAGS_HOST) $<
 
 #-------------------------------------------------------------------------------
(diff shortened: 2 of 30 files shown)


phil-blain:
The Travis build failed because the Makefile macro $(SCC) was not defined in the `Macros.travisCI_gnu` file. I think this file is a little outdated, as the macros SCC and SFC were there but were empty. I updated that. 
Also, the Travis build was using the clang preprocessor, I think this is an historical artifact so I removed that and set CPP to /usr/bin/cpp instead in the Macro file. I also added gcc to the travis.yml file.
If I'm out of line with these changes I will revert them.

phil-blain:
I updated all machine files to define correctly `SCC, SFC, CC, FC, LD`. I removed the references to MPI since it is not used in Icepack.

anders-dc:
Hi Phil, I used `clang` because it worked with some preprocessing directives where `gcc` did not. If these issues have been sorted out that's great. If you'd like, the Travis tests can be run in parallel with both `gcc` and `clang`, just to check beyond a single compiler.

phil-blain:
I agree that the external dependency generator is unlikely to change, but you are right that for consistency we can use $(DEPGEN) as the target.

phil-blain:
I just thought that the new make variable CFLAG_HOST should be documented. I will update the doc. Commit upcoming.

phil-blain:
I added a section to the doc. It can be previewed here : 
[https://icepack-phil-blain.readthedocs.io/en/build-system-robustness/user_guide/ug_running.html#cross-compiling](https://icepack-phil-blain.readthedocs.io/en/build-system-robustness/user_guide/ug_running.html#cross-compiling)

Issue #248, opened 2019-01-04
Add water isotopes


dabail10:
We have these in our version of CICE5. So, they would need to be added to both Icepack and CICE. Now that I have a working version of CICE6 in CESM, I am getting started on this.

apcraig:
Just to note, the first PR for water isotopes is #275 and https://github.com/CICE-Consortium/CICE/pull/423.  The main remaining issue is isotope conservation validation (could be science, technical, or diagnostics - TBD).  I will leave this issue open.

eclare108213:
See also #306, for which the isotope conservation issue can be a test case.

apcraig:
Closing, seems to be done.

PR #275, opened 2019-09-05
Isotope

dabail10:
#91  PR checklist
This PR adds the water isotope tracers to icepack. Comes from the CESM code.
- [X] Developer(s): 
 David Bailey (dabail10)
 https://github.com/CICE-Consortium/Test-Results/wiki/64a2cee85b.izumi.nag.190904.221343
https://github.com/CICE-Consortium/Test-Results/wiki/64a2cee85b.izumi.intel.190904.221343
https://github.com/CICE-Consortium/Test-Results/wiki/a5449f5207.izumi.gnu.190905.155549
https://github.com/CICE-Consortium/Test-Results/wiki/a5449f5207.izumi.pgi.190905.200002
    - [X] bit for bit
    - [ ] more substantial 
    - [x] Yes
    - [ ] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [X] Yes
        - [ ] No 
This includes the water isotope tracers. They are very similar to the aerosol tracers, but not like the zaero tracers. I would eventually like to treat the isotopes the same way as the zaero tracers, but we need the current implementation for the CESM. Results are bfb with tr_iso = .false., but fail the restart comparison with tr_iso = .true. This is because the isotopes are needed in the restart files. The ice_diag ascii files are identical. This will address #248


apcraig:
One issue is that it looks like at least one (maybe more) interface (icepack_atm_boundary) is modified which will break CICE until CICE is updated and tested.  If we push this to master, we cannot test any icepack changes in CICE and icepack master will be broken with respect to CICE.  I think my preference would be that CICE modifications be carried out before we merge this PR.  This will ensure that the new icepack mods are working in CICE and it will not block other changes to icepack moving forward.

This is one area where the current icepack interfaces are not that great, extensibility.  And by the way, I still don't know how/if we should proceed with any icepack interface refactor at this point.  But that's a separate issue.

dabail10:
Excellent point. I will modify the above that this requires a CICE PR as well.

apcraig:
You can ignore this comment, just testing something.

apcraig:
@dabail10.  Even if this is not quite ready for usage, I still have the same question as in September.  If we merge this now/soon, will be unable to update CICE with new Icepacks due to interface changes?  Can we have a CICE PR that addresses this issue?  We do not want to make the Icepack master unusuable in CICE for any extended period of time.  It's fine if it's a couple days and we have a plan to get past it.  But I worry that's not the case here.

dabail10:
I am doing extensive testing with both Icepack and CICE updated to master. This should be ready soon.

eclare108213:
This PR is being replaced by #307, so closing it now.

PR #281, opened 2019-10-31
Implement Floe Size Distribution (FSD) in Icepack

eclare108213:

## PR checklist
Implement floe size distribution in Icepack
- [x] Developer(s): 
    @lettie-roach and @eclare108213 
118 of 130 tests PASSED
9 of 130 tests FAILED
0 of 130 tests PENDING
https://github.com/CICE-Consortium/Test-Results/wiki/9b5a5670e4.badger.intel.19-10-31.041919
This includes FSD tests not available in the original code, so there are some 'misses'.  The failures are all expected -- the code is set to abort if the FSD and BGC are both turned on, until results from the combination can be looked at and fixed as needed.
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No but CICE is being modified in order to use this FSD code
    - [x] Yes and it could use more
    - [ ] No
    - [x] Yes but it is probably incomplete, needs a thorough review
        - [ ] Yes
        - [ ] No 

The current Icepack tests do not exercise wave breaking, which can be tested in CICE; a simpler form might be included in Icepack in the (near?) future.  Results from the current tests need closer scrutiny to make sure they are behaving as expected.

The default value of ice_ic is changed in icepack_in, to make the functionality the same as in CICE for the FSD.  This does not matter for general icepack initialization because each of the 4 grid cells is initialized manually, and ice_ic is only meaningful for restart runs.

There are a few more things to do, before the FSD is ready for use in CICE.  Most of what is needed for Icepack is here, but needing review and additional testing.  In particular:
- [x] @lettie-roach may make a few more changes
- [x] check documentation text and tables against actual code changes to make sure we caught everything
- [x] run all of the test suites on various platforms, compilers, etc
- [x] run the complete collection of tests with -s fsd12 to see what breaks. (Can this be done on the command line?)  BGC definitely will fail, and we might want to add other aborts, depending on the case.  
- [x] run the complete collection of tests with -s fsd1 and compare with an Icepack 1.1.2 baseline to check BFB (All tests pass but it is NOT bit-for-bit with current master, do we need to look into this further?, see https://github.com/CICE-Consortium/Icepack/pull/281#issuecomment-557596800)

See also the FSD [project board](https://github.com/CICE-Consortium/CICE/projects/3).


apcraig:
That travis testing is showing some failed tests for the bgc cases.  Are all bgc cases expected to fail right now and is that going to be fixed before this PR is merged?  Is fsd a feature that will always be on moving forward or is it something that gets turned on as an option? 

I will run a full test suite on conrad and izumi on multiple compilers now.

We should be able to add -s fsd12 to a test suite, and it should do what you expect.  We'll have to confirm that when we try.

eclare108213:
The BGC tests should fail only when FSD is turned on, because I added an abort if both FSD and BGC are turned on -- there are some changes that need to be made within the BGC routines to properly account for the FSD.  These can be fixed at some point in the future, but it'll take some work that will not be done in time for this PR; the most obvious approach, which I tried earlier this year, fails.  FSD will not be turned on by default until we have more experience with it; there is also some more development planned for it, especially as the new ridging scheme is implemented.  If the BGC tests are failing without FSD turned on, then we have a problem...  I'll look at that.  Thanks for testing, @apcraig.

apcraig:
It looks like the test reporting is a little broken.  A couple changes were made a couple weeks ago.  If you drill down on the test results for either badger or izumi (#fab7e3c02b), you'll see that the bgc tests are failing on all platforms.  I will look into the test reporting error.

eclare108213:
The BGC on / FSD off runs were failing (at least in part) because tr_fsd wasn't initialized.  I have fixed that.

apcraig:
@eclare108213 , could you push one more scripts modification.  In icepack/configuration/scripts/tests/report_results.csh at line 172 there are two else statements back-to-back.  Could you remove one of those and push that to your branch.  I will then retest with your latest fix to see if bgc is passing again on izumi and conrad.   travis is passing, so I think we probably are good.

apcraig:
Latest results for izumi are posted.  There are two failures with gnu, the fsd1 and fsd12 tests.  Those tests pass for pgi, intel, and nag.  The traceback for both izumi_gnu_smoke_col_1x1_debug_fsd12_run1year and izumi_gnu_smoke_col_1x1_debug_fsd12_run1year is

> Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
> 
> Backtrace for this error:
> #0  0x2b58e631327f in ???
> 
>
> #1  0x4da36b in __icepack_fsd_MOD_icepack_cleanup_fsd
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/columnphysics/icepack_fsd.F90:359
> #2  0x4d625c in __icepack_fsd_MOD_fsd_add_new_ice
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/columnphysics/icepack_fsd.F90:702
> #3  0x5840ba in __icepack_therm_itd_MOD_add_new_ice
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/columnphysics/icepack_therm_itd.F90:1739
> #4  0x57a13e in __icepack_therm_itd_MOD_icepack_step_therm2
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/columnphysics/icepack_therm_itd.F90:2058
> #5  0x474c2a in __icedrv_step_MOD_step_therm2
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/configuration/driver/icedrv_step.F90:414
> #6  0x405a75 in __icedrv_runmod_MOD_ice_step
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/configuration/driver/icedrv_RunMod.F90:155
> #7  0x405e8d in __icedrv_runmod_MOD_icedrv_run
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/configuration/driver/icedrv_RunMod.F90:56
> #8  0x402d5b in icedrv
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/configuration/driver/icedrv_MAIN.F90:48
> #9  0x402e52 in main
> 	at /home/tcraig/cice-consortium/icepack.lr.ifsd/configuration/driver/icedrv_MAIN.F90:28
> 

I will look into this as I suspect it may not be easily reproducible on other machines.  It seems to be gnu specific.

apcraig:
Just had a look at the refactor of the fsd cleanup.  Makes sense to me and looks good.  Will retest now.

apcraig:
This now has conflicts due to the changes in the icepack calls in the driver.  If you want some help resolving these conflicts, let me know.  We'll want any new arguments to icepack to be implemented with keyword=value approach.

eclare108213:
@apcraig if you have time, Yes, please fix the conflicts.  I won't be able to get to this until Thursday at the absolute earliest.  Thanks!

apcraig:
@eclare108213 , I'll update the interfaces in the next day.

apcraig:
I just updated to master, updated the interfaces, and resolved conflicts.  I ran quicksuite on izumi with 4 compilers.  I will next submit a full test suite on izumi.

apcraig:
The full izumi test suite results are here, #4b7abae372997
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks

eclare108213:
Thanks @apcraig.  Do you know what is causing the failures on intel?  Looks like some of them are cases with fsd turned off.

apcraig:
The jobs that failed with fsd off timed out.  I've seen that before on izumi.  The only jobs that actually fail are the fsd tests.  I probably need to update the runlength for izumi to minimize those kinds of failures.

I'm still unclear what still needs to be done.   For the fsd tests, I get

> FAIL izumi_nag_smoke_col_1x1_debug_fsd12_run1year test
> PASS izumi_nag_smoke_col_1x1_debug_fsd1_run1year test
> PASS izumi_nag_restart_col_1x1_fsd12 test 
> FAIL izumi_intel_smoke_col_1x1_debug_fsd12_run1year test
> FAIL izumi_intel_smoke_col_1x1_debug_fsd1_run1year test
> PASS izumi_intel_restart_col_1x1_fsd12 test 
> PASS izumi_pgi_smoke_col_1x1_debug_fsd12_run1year test
> PASS izumi_pgi_smoke_col_1x1_debug_fsd1_run1year test
> PASS izumi_pgi_restart_col_1x1_fsd12 test 
> FAIL izumi_gnu_smoke_col_1x1_debug_fsd12_run1year test
> PASS izumi_gnu_smoke_col_1x1_debug_fsd1_run1year test
> PASS izumi_gnu_restart_col_1x1_fsd12 test 

I don't think I've added a dataset.  Is there a dataset we need to add and should I test that?  I think we had some discussion about filenames and paths and such, but lost track.  I can test that on izumi or elsewhere if it's clear what we want to do.  There was also some testing to do with -s fsd12 and do we want to compare that with and without that flag?  I know I've done some of that before.  Let me know how I can help.  I'm happy to help update code and/or carry out testing.

phil-blain:
Minor point: I now this is not ready to merge yet, but before it is the title of the PR should be changed to something more meaningful like "Implement flow size distribution", in order to avoid having "Ifds1" as the commit message when we merge.

eclare108213:
@apcraig,  a few remaining questions:

Do tests on Izumi still fail, or were those just needing a longer allocation?  It sounds like the failures were real, and it's intel so maybe I can reproduce them on badger.  I'll try that.

I remember you'd done some testing with -s fsd options, and it seems like the results were fine, but I'd like to be sure.  There were two sets in the initial PR entry here, 
- running the entire test suite with -s fsd12 turned on to see what breaks (you did this -- what's the outcome?), and 
- running the complete collection of tests with -s fsd1 and comparing with an Icepack 1.1.2 baseline to check BFB.   I don't know whether to expect BFB, but it would be nice if it worked.

There is a wave-forcing file for FSD testing, but it's only available for gx3, and we'll add it with the CICE FSD PR, not here.  Icepack uses internally (driver) defined default forcing, only intended for testing.

eclare108213:
102 of 102 tests PASSED
0 of 102 tests FAILED
0 of 102 tests PENDING

I think this Icepack FSD PR is ready to merge, once we decide what to do about the interfaces.  I'm working on the associated CICE PR.

apcraig:
I will checkout the latest and run a suite of tests again.

apcraig:
For the current ifsd1 branch in icepack, hash aa7f82b3b7

on izumi, 4 compilers, full test suite, branch PASSES all tests and is bit-for-bit.

on gordon, 4 compilers, full test suite, branch PASSES all tests and is bit-for-bit except
  FAIL gordon_intel_restart_col_1x1_dyn compare icepack.68bc397d7b.191110-055308 different-data
    comparing the log files, only the "ice free" case is different.  I'm not sure why just this case on this machine+compiler fails when the other 7 (4 x izumi + 3 x gordon) pass.

on izumi, 4 compilers, full test suite, add -fsd12, branch PASSES all tests except all bgc tests fail

on izumi, 4 compilers, full test suite, add -fsd1, branch PASSES all tests except all bgc tests fail.  Also, NO test results are bit-for-bit with the current master.  In this case, the only difference is tr_fsd is on.  Two questions
  - do the non-bit-for-bit results need more review?  Picking out 1 case, of the 4 runs, only the "land" case seems to be the same.  Maybe this needs to be further investigated to make sure nothing is wrong?
  - should the bgc tests fail with fsd1?

All results can be seen here, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks under hash aa7f82b3b7d7c630.  There are 16 sets of results. The last 8 are the standard izumi and gordon test suites.  The next 4 from the bottom are tests set with -fsd12 (bgc tests fail and no comparisons are done).  The next 4 from the bottom are test set with tr_fsd=.true. set manually for all tests and all tests pass except bgc plus all comparisons fail versus the master.

apcraig:
I just resolved a conflict in this branch interactively in github resulting from PR #284.  You can see that diff here,

https://github.com/CICE-Consortium/Icepack/pull/281/commits/9b4af6b835da627a067e2c236203ad4a213f23da

eclare108213:
Reminder to self:  there is at least one subroutine name in the fsd code that starts icepack_, which is not an interface routine.  Fix that.

apcraig:
> Reminder to self: there is at least one subroutine name in the fsd code that starts icepack_, which is not an interface routine. Fix that.

Personally, I think all the subroutines in icepack should start with icepack_ and be named such that you could even tell which module/file they're in.  icepack_step_radiation -> icepack_shortwave_step_radiation and run_dEdd -> icepack_shortwave_run_dEdd.  I think that's not necessary or practical at this point, but I don't have a problem with non-public subroutines starting with icepack_.

eclare108213:
Thanks for the reminder - it would be good to change the fsd arguments to be optional if we are going to impose that on new code going forward.  I guess I was thinking that would be part of the icepack interface upgrade.   Looking through the code, though, this would create a LOT of "if(present())" checks, which doesn't seem like a good idea.  Would the optional variables only be in the icepack interface, and if they are present then load their values into non-optional arguments passed into lower-level subroutines (which would be ignored if tr_fsd=F)?  Or assume that if tr_fsd is true, then the optional arguments are indeed passed?

I just noticed a bug in line 1069 of ice_therm_itd.F90:  it should be rhoi*vicen+rhos*vsnon

apcraig:
The optional argument implementation questions are big ones.  We certainly would like to move toward optional arguments in the public interfaces.  The way this is most easily handled is to move toward module data within Icepack and move away from arguments where possible.  But that creates other tradeoffs.  Sometime before, I noted a few ways to handle the optional arguments.  1) move to module data where the flags we have trigger different features and move away from arguments inside Icepack.  2) lots of "if present" followed by lots of calls to the same interfaces with different arguments.  3) present checks in the public interfaces that instantiate dummy arguments to pass further down the calling tree as needed.  None are ideal.  There are other options too like refactor the code so different features are implemented via their own interfaces.  We are partly in this situation because the Icepack library came out of a working code instead of being designed as a library itself from the start.  The more I look at and understand the Icepack code, the less sure I am that we can overcome that without a lot of work.  I think it's fine to leave the fsd arguments as is for now and look at making them optional as part of the current interface refactor effort.

Is the ice_therm_itd bug an answer changer?  If so, I recommend we make that a separate and clean PR, so the fsd and interface refactors both can be tested via bit-for-bit comparisons more easily.

apcraig:
Testing looks good.  Full test suites with 4 compilers on each of gordon and izumi pass.  See hash 9b4af6b835 at https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks.  One test failed on izumi intel which I believe is not something to be concerned about.  The comparison fails are for fsd tests where there are no baselines.

eclare108213:
Please fix the bug first, noted in my previous comment.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
OK, I've fixed the bug in icepack_therm_itd in my sandbox and am retesting icepack.  If this is an answer changer, I still think we should consider fixing it as a separate PR though to preserve bit-for-bit in the fsd PR.

eclare108213:
Okay, let’s continue to discuss the interface implementation and go ahead with FSD. The bug was introduced with my FSD changes and so should be fixed in this PR.    I think it only affects a coupling field so wouldn’t show up as answer changing in stand-alone runs.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
Sounds good @eclare108213 thanks.

eclare108213:
Also, please check the to-do list at the top of this PR.  I think it’s all done but want to confirm. Seems like there was still a box left unchecked.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
The box left unchecked was testing with -s fsd1 vs current master for bit-for-bit.  I added a comment that all tests pass with -s fsd1 but it is NOT bit-for-bit with master.  Do we need to look into this?

eclare108213:
No, I wouldn’t expect it to be BFB.  Miracle if it were!  I wonder if it’s worth running through QC as a sanity check, though.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
Is it not bit-for-bit because of order of operations or because a feature is actually turned on that changes answers?  I could do a qc test, will put that on my list.

eclare108213:
Essentially order of operations.  This test uses the FSD code instead of the hard-wired floe size in the original code, but is supposedly configured equivalently (a single, constant, 300 m floe size).  So QC should pass unless there is something wrong with the FSD code.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
Testing with the icepack_therm_itd fix went fine, everything passed and was bit-for-bit.  I have pushed that fix to the branch.

apcraig:
I ran a qc test with the current fsd cice branch (3b487bde1bd12) and the current icepack branch (cf1ecc0e14) which includes the fix to icepack_therm_itd.  I ran two qc tests, one with out-of-the-box settings and a second, identical case with tr_fsd=.true. set manually in ice_in.  The qc test seems to have failed,

> ./cice.t-test.py $WORKDIR/CICE_RUNS/gordon_intel_smoke_gx1_64x1_medium_qc.qcfsdb/ $WORKDIR/CICE_RUNS/gordon_intel_smoke_gx1_64x1_medium_qc.qcfsdt/
> INFO:__main__:Running QC test on the following directories:
> INFO:__main__:  /p/work1/apcraig/CICE_RUNS/gordon_intel_smoke_gx1_64x1_medium_qc.qcfsdb/
> INFO:__main__:  /p/work1/apcraig/CICE_RUNS/gordon_intel_smoke_gx1_64x1_medium_qc.qcfsdt/
> INFO:__main__:Number of files: 1825
> INFO:__main__:2 Stage Test Passed
> INFO:__main__:Quadratic Skill Test Failed for Northern Hemisphere
> INFO:__main__:Quadratic Skill Test Failed for Southern Hemisphere
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> INFO:__main__:
> ERROR:__main__:Quality Control Test FAILED
>

eclare108213:
That is disappointing.  I'll look into it.  First, could you check some things in the diagnostic output for the fsd run?  It should have tr_fsd=T, nfsd=1, wave_spec_type=none in namelist; it looks like defaults are set appropriately so these should be okay.  The other thing to check is the floe size, 'avg fsd rep radius' in the log file, to see whether it is approximately 300 m.  I don't know whether to expect it to stay exactly 300.  I'll work on this, hopefully Monday, but @apcraig if you could check the output, that would give me a head start.  Thanks for running the tests!
Looks like python is a pain...

eclare108213:
I had a quick look at an nfsd=1 run I'd done earlier, and it's showing the avg radius ~150 m.  This means the floe size bin isn't initialized right.  In icepack_fsd.F90, try changing
         lims = (/ 6.65000000e-02,   3.0e+02 /)
to
         lims = (/ 6.65000000e-02,   5.999335e+02 /)
and see if that fixes it?  The upper bound here is not exactly 600 m because the lower bound is not exactly 0.  It looks like the calculation will not keep the floe size radius at exactly 300 m, regardless, but this ought to be close.   I'll look at history output when I get back to my office to see whether it's wonky.

eclare108213:
I take it back, the floe size diameter should be 300 m, so the radius should be 150 m.  The small difference due to the lower bound might make be making a difference for the QC -- I'm not sure how sensitive it is.  This could be tested:  in icepack_fsd.F90, try changing
      lims = (/ 6.65000000e-02, 3.0e+02 /)
to
      lims = (/ 6.65000000e-02, 2.999335e+02 /) ?
Or to
      lims = (/ 0.0, 3.0e+02 /) ?
I'll work on this when I get back.

apcraig:
I ran two additional qc tests with 

> lims = (/ 6.65000000e-02, 2.9335e+02 /) 
> and
> lims = (/ 0.0, 3.0e+02 /) ?

Running the t-test, all the cases with tr_fsd=.true. and different lims settings pass the qc test against each other (and are not bit-for-bit).  But none of the tr_fsd=.true. cases pass versus the case with tr_fsd=.false.   It seems like there is a difference in the two implementations (tr_fsd = false and true).  What part of the code should we be looking at?  Also, in the long term, we probably want just one implementation that handles either tr_fsd setting to reduce maintenance among other things.  If we have two, as either is upgraded, the other has to follow.

eclare108213:
@apcraig, thanks for running these tests.  I'll compare history output from CICE runs with and without tr_fsd on, tomorrow, to see whether/how they significantly differ, and maybe debug in Icepack.  I agree that we should have only one implementation for nfsd=1, and that should be the original (tr_fsd=F), since it's a lot cheaper to run.  But this test is a useful sanity check for the FSD code.   I'm glad we set it up!

eclare108213:
I have looked into the nfsd=1 case and now realize that the FSD operates very differently from the standard code, in that each thickness category is associated with a whole distribution of floe sizes.  This means that, e.g. when new frazil ice is added, it is distributed among all of the different ice thickness categories, while in the standard code, we add all of the new ice to the thinnest thickness category unless there's not room for the whole volume, in which case it's spread among categories.  So the calculations for tr_fsd=T / nfsd=1 are fundamentally different from tr_fsd=F.  It's reassuring that nfsd=1 maintains the floe diameter that it starts with, and I think we should consider this good enough for this PR.  I believe this PR is ready to be merged, pending final changes to and tests of the full [CICE FSD PR](https://github.com/CICE-Consortium/CICE/pull/382).

PR #290, opened 2019-12-09
report_results.csh: add html anchors to hashes and machines

phil-blain:
## PR checklist
    Add html anchors to hashes and machines in Test-Results wiki
- [x] Developer(s): 
    P. Blain
- [x] Suggest PR reviewers from list in the column to the right.@apcraig
    None
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 

The hashes and machine names inserted by the script in the different pages
of the Test-Results wiki are formatted as bold (`** ${hash} **:`).

Change this to level 4 HTML titles (\<h4\>) so that an HTML anchor is
automatically created, which makes it easier to link directly to specific
hash from a pull request.

### DIFF
--- a/configuration/scripts/tests/report_results.csh
+++ b/configuration/scripts/tests/report_results.csh
@@ -255,10 +255,11 @@ end
 #=====================
 
 set chk = 0
-if (-e ${hashfile}) set chk = `grep "\*\*${hash}" ${hashfile} | wc -l`
+if (-e ${hashfile}) set chk = `grep "#### ${hash}" ${hashfile} | wc -l`
 if ($chk == 0) then
+# Note: the line '#### ${hash}' is not a comment since it's in the here doc
 cat >! ${hashfile} << EOF
-**${hash}** :
+#### ${hash}
 
 | machine | compiler | version | date | test fail | comp fail | total |
 | ------ | ------ | ------ | ------  | ------ | ------ | ------ |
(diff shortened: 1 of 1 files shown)


phil-blain:
xref: CICE PR with the same changes: CICE-Consortium/CICE#388

phil-blain:
Thanks for the review :)
> First, no link is really created and I'm still unclear what the purpose of this is. 

The purpose of this PR is to create HTML anchors (the `#anchor` part at the end of a URL) on the pages of the Test-Results wiki, not links. This means that when we reference the result of a test in a related PR, we can link directly to the test hash instead of just mentioning it. So instead of writing as you did above:
>  See https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks hash #89af3098a.

We can simply link directly to the hash:
"See https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#89af3098a892d595967c0a690c4a37ace616d7cf"

The anchor-including URL of a specific test can be obtained by hovering on the hash on the test result page and then clicking on the anchor icon that appears (but since the hash is the anchor, we can also type "https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks", type '#' and then paste the full hash from somewhere else if it's more convenient. Both work.

>  Second, the results are not being grouped by hash anymore. I think the "chk" logic two lines above the "Note" needs to be modified to not check for "**${hash}" but to check for "#### ${hash}" for both by hash and by machine. 

That's true, I missed that. I will modify the grep argument and update both PRs. After that all results for the same hash should go in the same table. 

> Third, for the machines list, this will not be fully backwards compatible in the sense that it will create new machines tables.

That's also true. We could edit the machine pages manually, doing the same markup changes as in this PR (bold -> H4) and then any new test would not create new tables, as long as everybody posting results to the page updates their local repo(s) after the two PRs are merged.

apcraig:
Thanks @phil-blain, I understand the purpose the links serve now.  It's not to point to something outside the results page, this makes it easier to point to a specific section ON the results page.  I agree that's a nice feature, thanks for thinking of it and proposing it.  Once the suggested fixes are put in place, I am happy to merge this.  I'm not super worried about the backwards compatibility issue.  Once we start to generate some test results with this new feature, I'll plan to manually update a few pages to support the transition, but it's not a critical issue.   Thanks.

phil-blain:
I changed all instances of `\*\*` to `#### `
 (I had forgotten 2 on the first force-push).

PR #302, opened 2020-02-21
Add initialization of n_aero to icepack driver

apcraig:
## PR checklist
    Add initialization of n_aero in icepack driver
- [X] Developer(s): 
    apcraig
    Tested in icepack on conrad with 4 compilers.  All tests pass EXCEPT the pondcesm test case indicates non-bit-for-bit vs the baseline.  I have compared the ice_diag files for all pondcesm cases vs the baseline, and they are bit-for-bit, but the restart file (which is what is compared in the test) is not.  I suspect something is written slightly different into the restart file that does not affect answers.  This seems to happen just for the pondcesm case.  I don't understand why.  Because the restart files are binary, it's not easy to identify the differences.  But I think we can accept this since the ice_diag files are identical.
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#7259556232cebf37f4c9b42b4ba0b866a07d196b 
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

### DIFF
--- a/configuration/driver/icedrv_init.F90
+++ b/configuration/driver/icedrv_init.F90
@@ -674,6 +674,7 @@
          write(nu_diag,1020) 'nslyr   = ', nslyr
          write(nu_diag,1020) 'nblyr   = ', nblyr
          write(nu_diag,1020) 'nfsd    = ', nfsd
+         write(nu_diag,1020) 'n_aero  = ', n_aero
          write(nu_diag,*)' '
          write(nu_diag,1020) 'nx      = ', nx
          write(nu_diag,*)' '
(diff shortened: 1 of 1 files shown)


apcraig:
Thanks @dabail10 and @eclare108213.  In general, the values have to be set within the columnphysics part of icepack thru the icepack interfaces.  How any driver handles management in the driver is up to the driver.  You can store the variable in the driver and reuse them there or you can query icepack to get the value.  That is not specified.  The main issue is that things need to be set in the columnphysics part of icepack and you cannot access that data via a "use" statement.

I will merge the PR now and then it can be included in weekend testing.  Thanks!

PR #303, opened 2020-03-06
Update albedo initialization for identical order of operations

apcraig:
## PR checklist
    fix albedo initialization which resulted in one non-exact restart
- [X] Developer(s): 
    apcraig
    Tested on conrad full suite with 4 compilers, all pass, all bit-for-bit [(results)](https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#92d673e5ad3084b2aa07606e4875a3614cc675c1)
    Tested on conrad full suite with 4 compilers with iso1, all tests pass, a few tests fail regression due to compiler optimization related to iso1 change (not this PR) [results](https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#92d673e5ad3084b2aa07606e4875a3614cc675c1)
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

With the iso1 changes, one test was failing exact restart on one machine and compiler.  After some debugging, it turns out to be an order of operations issue in initialization of albedos that has been there forever but never resulted in problems with exact restart.  For whatever reason, this one test on one compiler on one machine picked it up finally.   It is fixed and the answers are also bit-for-bit with prior results.

I also fixed initialization of nt_bgc_S when solve_zsal is off.  It was not uniquely defined but this did not seem to matter to runs.  Still, it should be set.

I added some diagnostics for restart read and write to more easily confirm identical fields written and read.

I modified how the ice_diag files are saved to the log directory by attaching the run timestamp to the filenames.  This was long overdue.  Prior, runs in the same directory (including restart tests) just overwrote the files on each run.  We may want to eventually also change the way we save baselines or rename files in the run directory as this feature is only implemented in the copy of the ice_diag files to the case log directory.

I removed an OMP directive in the albedo initialization as it will not materially help performance since the code is run once.  It's also a trivially cheap piece of code.

### DIFF
--- a/configuration/driver/icedrv_restart.F90
+++ b/configuration/driver/icedrv_restart.F90
@@ -328,7 +328,7 @@
          work2              ! input array (real, 8-byte)
 
       real (kind=dbl_kind) :: &
-        minw, maxw          ! diagnostics
+        minw, maxw, sumw    ! diagnostics
 
       character(len=*), parameter :: subname='(read_restart_field)'
 
@@ -339,7 +339,8 @@
 
       minw = minval(work)
       maxw = maxval(work)
-      write(nu_diag,*) minw, maxw
+      sumw = sum(work)
+      write(nu_diag,*) subname, minw, maxw, sumw
       
       end subroutine read_restart_field
       
@@ -367,12 +368,20 @@
       real (kind=dbl_kind), dimension(nx) :: &
          work2             ! input array (real, 8-byte)
       
+      real (kind=dbl_kind) :: &
+        minw, maxw, sumw    ! diagnostics
+
       character(len=*), parameter :: subname='(write_restart_field)'
 
       do n = 1, ndim
         work2(:) = work(:,n)
         write(nu) (work2(i), i=1,nx)
       enddo
+      
+      minw = minval(work)
+      maxw = maxval(work)
+      sumw = sum(work)
+      write(nu_diag,*) subname, minw, maxw, sumw
       
       end subroutine write_restart_field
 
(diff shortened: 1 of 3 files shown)


apcraig:
It would be good to get this fix into the repo by Friday evening for weekend testing.  We may also want to update CICE prior to the weekend if we can as well.   Will see how reviews go.

configuration/driver/icedrv_restart.F90
```diff
@@ -339,7 +339,8 @@ subroutine read_restart_field(nu,work,ndim)
 
       minw = minval(work)
       maxw = maxval(work)
-      write(nu_diag,*) minw, maxw
+      sumw = sum(work)
+      write(nu_diag,*) subname, minw, maxw, sumw
```
dabail10: Do you still want the print statements here?

eclare108213: They were there before, and I've found them to be handy.

apcraig: These were there before and I just extended the feature a bit.  One of the first things I wanted to check was whether all the fields were read as written and whether they were identical.  It's helpful to have the same diagnostics written when writing a restart as when reading it, and the sum is a new diagnostic that provides something like a checksum value of the field.

apcraig:
Just one other comment.  The problem, in some ways, is that the same code is implemented twice.  I fixed it so they would compute fields with the same order of operations.  But the right thing to do would be to pull out this code and create a subroutine with it where both parts of the model are calling it.  This is all in the icepack driver, so it wouldn't affect icepack columnphysics.  I didn't do that as it's a bit more intrusive, I thought I'd just keep the fix as simple as possible.  But it's something that could or maybe should be done.

Issue #249, opened 2019-01-04
Atmospheric boundary layer loop convergence on ustar.


dabail10:
We have added some code based on Mahrt and Sun, 1995, MWR to increase the wind speed in very stable boundary layer regimes. This has a namelist flag, use_coldair_outbreak_mod and only affects icepack_atmo.F90 (atmo_boundary_layer).

We also have code that will stop the iteration when ustar is changing less than a specified convergence tolerance. This could potentially reduce the number of iterations in icepack_atmo.F90 (atmo_boundary_layer). This can be turned off and the fixed value of iterations is used.

We also have some code to only print the diagnostic message about high frequency RASM coupling once.

eclare108213:
Is the Mahrt and Sun option applied instead of the Jordan option that tries to address the same issue?  If so, then maybe the option should be generalized.  If not and it's applied on top, well, maybe we should rethink that.

dabail10:
It is applied on top of the Jordan et al. option and only in regions where the temperature difference between the surface and the lower atmosphere is above a particular threshold.

proteanplanet:
Please assign me to this issue.  I will be testing it in E3SM.

proteanplanet:
Hi Dave,  Please can you point us to some documentation on this change.  Has it been written up in a paper as part of CESM V2?  This change will affect other coupled models, and it needs to be consistent with their boundary layers, too, so will most likely need to be implemented as a CESM option. As Elizabeth points out, it needs to be consistent with what is already in the boundary layer for CICE, hence the need for me to read up on the change. I've look at the paper, but I'm not seeing consistency with the existing CICE boundary layer method, probably due to a lack of intelligence on my part.  - Andrew

proteanplanet:
I see that  use_coldair_outbreak_mod  switches on one line or not, but that doesn't necessarily make it consistent with the rest of the boundary layer parameterization.   Perhaps you are referring to a discussion with Elizabeth not on this issues page?

dabail10:
I don't understand. We need it as a part of our CMIP runs. It is switchable and would be off in the default CICE model. So, no other groups have to use it. I am trying to keep our version up to date with the Consortium version.

eclare108213:
[...]

PR #315, opened 2020-04-22
Atmospheric boundary layer convergence on ustar.

dabail10:

## PR checklist
Add flux convergence criteria for ustar and cold air outbreaks code. This addresses issue #249.
- [X] Developer(s): 
dabail10 (David Bailey)
Test-Results.wiki/icepack_dev/b25d616c3e.izumi.intel.20-04-22.212638.0.md
Test-Results.wiki/icepack_dev/b25d616c3e.izumi.nag.20-04-22.212638.0.md
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
This code is bfb when turned off (default). They will change answers only when turned on. These are needed for the CESM. Note that I have also change atmo_boundary_layer and atmo_boundary_const from public to private. These are not called outside of icepack_atmo.F90. I have also updated the test length time for izumi to 50 minutes. This allows all of the base_suite tests to complete.


doc/source/master_list.bib
```diff
@@ -47,6 +47,7 @@ @string{BAMS
 @string{BLM = {Boundary-Layer Meteorol.}}
 @string{AG = {Ann. Glaciol.}}
 @string{PHYS = {Physica D}}
+@string{JAMES = {J. Adv. Modeling the Earth Sys.}}
```
duvivier, line 50: It looks like Lettie's 2019 paper is JAMES too. Would you mind fixing hers like you did for Brady et al. 2019?

dabail10:
I was guessing you were going to want name changes, so I was waiting for your suggestions. :)

apcraig:
I had a look at the code and nothing jumps out at me.  I generally think it's nice for "use" statements to appear at the module level rather than the subroutine level for easy reuse throughout the module, but we do not currently require that.

Regarding the idea of public subroutines in icepack.  I think we want to be very careful how we proceed.  It seems to me that a subroutine is public in icepack if it's needed in a different module in icepack.  Otherwise, the subroutine should not be public.  In addition, any subroutine that is exposed to the outside, should be public via the icepack_intfc module.  If there are people that are using icepack in different ways then we should address that generally.  If someone needs an interface exposed, we should expose it thru the icepack_intfc module just like all other public interfaces.  There should be no back doors into the icepack interfaces for special uses.  The community can request those kinds of changes to the public interfaces easily and we can respond quickly.  Separately, if someone is using a subset of icepack code, pulling out subroutines, not using the icepack_intfc layer, and so forth, then I think they are largely on their own in terms of managing the "use".  So, I don't think there is any case where we should allow an interface to be public unless it's used in another module or it's use in the icepack_intfc module in particular.  I propose we continue to proceed down that path including the changes that are included in this PR.  If we need to expose more interfaces via the icepack_intfc module, we can.  But I suggest we avoid exposing interfaces ad hoc especially if we're not sure why.

eclare108213:
I agree.  By making these subroutines not public, we might find out who needed them before, or maybe too many years have passed and this issue is moot.

dabail10:
After discussing this further with the Atmospheric Model Working Group, I am going to hold off on the cold air outbreak code. We are not likely going to use it in CESM3, and so there is no need to add it to CICE6.

dabail10:
Updated test results just for the natmiter_conv flag:

Test-Results.wiki/icepack_dev/91a0f2f043.izumi.intel.20-04-27.163330.0.md
Test-Results.wiki/icepack_dev/91a0f2f043.izumi.nag.20-04-27.163330.0.md

apcraig:
It looks like this code is probably ready to merge.  @dabail10, anything else outstanding?  @eclare108213, the cold air outbreak feature has been removed, does the code look OK to you?

columnphysics/icepack_atmo.F90
```diff
@@ -271,37 +267,43 @@ psixh     ! stability function at zlvl   (heat and water)
...
-                + (c1 - stable)*psimhu(xqq)
-         psixh  = psimhs*stable &
-                + (c1 - stable)*psixhu(xqq)
-
-         ! shift all coeffs to measurement height and stability
-         rd = rdn / (c1+rdn/vonkar*(alz-psimh))
-         rh = rhn / (c1+rhn/vonkar*(alz-psixh))
-         re = ren / (c1+ren/vonkar*(alz-psixh))
+         if (.not. (natmiter_conv > c0 .and. &
+                 abs(ustar - ustar_prev)/ustar <= natmiter_conv)) then
```
eclare108213, line 275: A couple questions re lines 270-275 in icepack_atmo.F90:
The value of ustar_prev being compared with seems a little arbitrary.  Why 2*ustar?
It looks like natmiter_conv does not change during the loop, so it would be more efficient to evaluate that part of the conditional outside of the loop.  Does that make sense?

dabail10: This is just to initialize the convergence loop and make sure it does it at least once.

eclare108213: Is it ever possible for ustar=0 on the first iteration, or thereafter?
Is ustar ever negative? 
For efficiency, could we remove the check on natmiter_conv and rewrite the lines 272-275 as
`      do k = 1, natmiter`     
`             if (abs((ustar - ustar_prev)/ustar) > natmiter_conv)) then`
?
Also, now that I'm looking at this, I don't like the 'n' at the beginning of natmiter_conv.  That denotes a number (i.e. integer, the number of iterations, in this case) but the _conv variable is a tolerance rather than an integer.  It's only a little confusing to those of us (maybe only me) who knows the convention for variable naming in CICE, which was inherited from POP...

duvivier: Ustar (friction velocity) in the atmosphere could in theory be zero but it can never be negative.

proteanplanet: ustar as it is currently set cannot be zero.  It is dependent on the minimum allowable wind speed, umin, which is greater than zero in the initialization of atmo_boundary_layer:

 185       if (highfreq) then
 186        umin  = p5 ! minumum allowable wind-ice speed difference of      0.5 m/s
 187       else
 188        umin  = c1 ! minumum allowable wind speed of 1m/s
 189       endif

This limit is to ensure that sensible and latent heat flux are not zero when the true wind magnitude are zero - if umin is ever set to zero, this breaks the science as well as the code.  Wind stress magnitude can still be zero, but that's performed separately.

proteanplanet: > Also, now that I'm looking at this, I don't like the 'n' at the beginning of natmiter_conv. That denotes a number (i.e. integer, the number of iterations, in this case) but the _conv variable is a tolerance rather than an integer. 

I agree with Elizabeth on this.

eclare108213: Thanks for taking a look at this, @proteanplanet.
@dabail10 it looks like there are 2 things remaining to address:  removing n from natmiter_conv, and changing the conditional as noted above to (potentially) make the loop more efficient.  For the latter, let's make sure that my logic is correct...

dabail10:
I have made the requested changes. Everything is still bfb with the default settings.

proteanplanet:
I have a further question about this change before I approve the PR.  The default parameter setting for natmiter is 5, athough it can also be set in the namelist.  This comes from CESM. In RASM, it is set to 25 to improve convergence.   I can understand a convergence test for 25 iterations, but not 5.    Would the consoritium be open to increasing the default value of natmiter now that it includes a convergence test?  What is the sensitivity to natmiter in CESM and stand-alone CICE?

proteanplanet:
Also, I thought @eclare108213 's suggested change was to improve efficiency in the convergence so that the loop broke once convergence had been met, with either a break or while command.  Is that not feasible?  What happens when natmiter is set to a very high number, say, for argument's sake, 500?

dabail10:
Good questions Andrew. I'm not sure why a "while" loop was not used here originally. I didn't that was what Elizabeth was suggesting above though? I think a "break" is kind of akin to a goto statement and is frowned on. However, if people feel strongly about this, I can do it. I think the default for natmiter should remain at 5, but allow users to change it as needed. This was more to standardize the computation in the CESM coupler for ocean-atmosphere fluxes and the computation in CICE.

proteanplanet:
Thanks for doing this. With this last change, I think this would be an improvement over everything we've previously had in the flux calculations.  

A while loop would leave this less to chance on any particular compiler optimization.   I should have brought this up first time around and I perhaps misunderstood Elizabeth's original comments about efficiency.

eclare108213:
When I first made the suggestion, I was thinking that it would jump out of the loop, but I didn't follow up on that.  I think that was coded before I knew about 'while'.  Actually, I think that code originally came from CSM!  Or maybe PCM.   :)   It's old.  A while implementation would be nicer.

proteanplanet:
Thanks for doing this. With this last change, I think this would be an improvement over everything we've previously had in the flux calculations.  

A while loop would leave this less to chance on any particular compiler optimization.   I should have brought this up first time around and I perhaps misunderstood Elizabeth's original comments about efficiency.

dabail10:
Ok. Coded as a while loop. The tests are still bfb with the default options.

PR #326, opened 2020-07-07
Shortwave redistribution option added to namelist.

dabail10:

## PR checklist
We need the "shortwave redistribution" option for ktherm=2 for CESM. I believe it was needed for E3SM as well. So, I have added it here, but with an option to turn
it off via the namelist.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_mach_forks#izumi
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
This is bfb with sw_redist=.true. and ktherm=1 and sw_redist = .false. and ktherm=2. Since the default for sw_redist will be .false. we should warn people who use the BL99 thermo (ktherm=1) that this will now be off and they will need to turn it on if necessary for their runs.


apcraig:
Looks like some hardwired parameters were moved to namelist and the results are bit-for-bit.    Why are the alt03, alt04, and thermo1 test cases bit-for-bit with the code changes?

dabail10:
For those tests, I added sw_redist = .true.

dabail10:
This also helps get rid of some CESMCOUPLED ifdefs.

doc/source/science_guide/sg_thermo.rst
```diff
@@ -50,6 +50,12 @@ the :ref:`thermo-growth` section. We begin by describing the melt ponds and sur
 forcing parameterizations, which are closely related to the ice and snow
 surface temperatures.
 
+Sometimes instabilities can arise when the temperature is close to the melt point in
+the snow and sea ice and there is abundant internal shorwave absorbed. One can chose
```
eclare108213, line 54: typo:  chose should be choose

apcraig:
> For those tests, I added sw_redist = .true.

I'm still a little confused.  I understand that in alt03, alt04, and thermo1, sw_redist is now true.  What I don't understand is why everything is bit-for-bit with the prior results when sw_redist is true for some cases and false for other cases.  Was another flag (which one?) playing the same role as sw_redist prior to this change?  

As a followup, subroutine temperature_changes_salinity has some new code that is on when sw_redist is true.  Is this just diagnostic?

dabail10:
The old code was only present in ice_therm_bl99.F90 (ktherm=1). It was not present in ice_therm_mushy.F90 (ktherm=2). So, it was only effectively "true" for the ktherm=1 case.

doc/source/science_guide/sg_thermo.rst
dabail10: Fixed.

apcraig:
I think all comments have been addressed and this is ready to merge.

PR #307, opened 2020-03-26
Isotopes for Icepack

eclare108213:
This PR is not yet ready to be merged - further review, testing and documentation is required.

- [x] Short (1 sentence) summary of your PR: 
Addition of isotope diagnostics following @dabail10's implementation for CESM.
- [x] Developer(s): 
    @dabail10 @eclare108213 @apcraig 
- [x] Suggest PR reviewers from list in the column to the right.
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
Initial tests by @apcraig looked okay
(see https://github.com/eclare108213/Icepack/pull/1) 
but this needs to be reconfirmed.
- How much do the PR code changes differ from the unmodified code? 
    - [x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on CICE or any other models?
    - [ ] Yes
    - [x] No
- Does this PR add any new test cases?
    - [x] Yes
    - [ ] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/.)
    - [x] Yes
    - [ ] No, does the documentation need to be updated at a later time?
        - [x] Yes
        - [ ] No 
Documentation for the interfaces is currently here but the isotope physics and code description needs to be added.
- [x] Please provide any additional information or relevant details below:

I will add the associated CICE PR shortly.  Both this PR and the CICE one need @dabail10 's scrutiny to make sure the results are correct.

PR #329, opened 2020-07-23
Update CPPs and Documentation

apcraig:
## PR checklist
    Update CPPs and Documentation
- [X] Developer(s): 
    apcraig
    Ran quick suite on 4 compilers on gordon.  Should not affect standalone Icepack.  https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#8efeb5970607894ee4286684680f4acb57614ae8
   Also tested in CICE with https://github.com/CICE-Consortium/CICE/pull/490
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

### DIFF
--- a/columnphysics/icepack_isotope.F90
+++ b/columnphysics/icepack_isotope.F90
@@ -482,10 +482,6 @@
 
       wiso_alpi = exp(alpai(isp)/tk**2 + alpbi(isp)/tk + alpci(isp))
 
-#ifdef NOFRAC
-      wiso_alpi = c1
-#endif
-!
       return
       end function wiso_alpi
 
--- a/columnphysics/icepack_kinds.F90
+++ b/columnphysics/icepack_kinds.F90
@@ -5,6 +5,7 @@
 !
 ! author: Elizabeth C. Hunke and William H. Lipscomb, LANL
 ! 2006: ECH converted to free source form (F90)
+! 2020: TC added support for NO_I8 and NO_R16
 
       module icepack_kinds
 
@@ -17,10 +18,18 @@
                             char_len_long  = 256, &
                             log_kind  = kind(.true.), &
                             int_kind  = selected_int_kind(6), &
+#if defined (NO_I8)
+                            int8_kind = selected_int_kind(6), &
+#else
                             int8_kind = selected_int_kind(13), &
+#endif
                             real_kind = selected_real_kind(6), &
                             dbl_kind  = selected_real_kind(13), &
+#if defined (NO_R16)
+                            r16_kind  = selected_real_kind(13)
+#else
                             r16_kind  = selected_real_kind(33,4931)
+#endif
 
 !=======================================================================
 
--- a/doc/source/user_guide/ug_case_settings.rst
+++ b/doc/source/user_guide/ug_case_settings.rst
@@ -9,30 +9,30 @@ There are two important files that define the case, **icepack.settings** and
 **icepack_in**.  **icepack.settings** is a list of env variables that define many
 values used to setup, build and run the case.  **icepack_in** is the input namelist file
 for the icepack driver.  Variables in both files are described below.  In addition,
-the first table documents available C Preprocessor Directives.
+the first table documents available C Preprocessor Macros.
 
 .. _tabcpps:
 
-Table of C Preprocessor (CPP) Directives
+Table of C Preprocessor (CPP) Macros
 ---------------------------------------------------
 
-The Icepack model supports a few C Preprocessor (CPP) Directives.  These
+The Icepack model supports a few C Preprocessor (CPP) Macros.  These
 can be turned on during compilation to activate different pieces of source
 code.  The main purpose is to introduce build-time code modifications to
 include or exclude certain libraries or Fortran language features, in part to
 support CICE or other applications.  More information
 can be found in :ref:`cicecpps`.  The following CPPs are available.
 
-.. csv-table:: **CPP directives**
+.. csv-table:: **CPP Macros**
    :header: "CPP name", "description"
    :widths: 15, 60
 
    "",""
-   "**General Directives**", ""
+   "**General Macros**", ""
    "NO_I8", "Converts integer*8 to integer*4."
    "NO_R16", "Converts real*16 to real*8."
    "",""
-   "**Application Directives**", ""
+   "**Application Macros**", ""
    "CESMCOUPLED", "Turns on code changes for the CESM coupled application                          "
    "CICE_IN_NEMO", "Turns on code changes for coupling in the NEMO ocean model"
 
--- a/doc/source/user_guide/ug_case_settings.rst
+++ b/doc/source/user_guide/ug_case_settings.rst
@@ -29,7 +29,6 @@ can be found in :ref:`cicecpps`.  The following CPPs are available.
 
    "",""
    "**General Directives**", ""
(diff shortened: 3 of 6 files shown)


doc/source/user_guide/ug_running.rst
```diff
@@ -201,6 +201,25 @@ To add some optional settings, one might do::
...
+C Preprocessing (CPP) Directives
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are a few C Preprocessing Directives supported in the Icepack model.  These
+support certain coding features to be excluded or included during the compile.  They
+exist in part to support the CICE model and other applications that use Icepack.
+
+For standalone Icepack, The CPPs are defined by the `CPPDEFS` variable in the Icepack
+Makefile.  They are defined
+by passing the -D[CPP] to the C and Fortran compilers (ie. -DNO_I8) and this
```
eclare108213, line 216: Instead of 'ie.' please use 'e.g.'
I don't know the Latin, but I think i.e. means "that is," and e.g. means "for example,"

apcraig:
I agree the NOFRAC should be moved to namelist.  I was thinking that might take some coordination with CICE and think it will, but I believe the Icepack upgrade can be done first.  I'll update this PR with that change.  Good idea.

apcraig:
The NOFRAC cpp has been removed.  The quick suite was run with 4 compilers on gordon and all results are bit-for-bit.

columnphysics/icepack_kinds.F90
```diff
@@ -17,10 +18,18 @@ module icepack_kinds
...
+#else
                             int8_kind = selected_int_kind(13), &
+#endif
                             real_kind = selected_real_kind(6), &
                             dbl_kind  = selected_real_kind(13), &
+#if defined (NO_R16)
+                            r16_kind  = selected_real_kind(13)
+#else
                             r16_kind  = selected_real_kind(33,4931)
+#endif
```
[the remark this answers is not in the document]
apcraig: The NO_I8 and NO_R16 are definitely compiler dependent which means they may be required on certain machines.  And this could certainly change.  For instance, NO_R16 is defined for a few machine/compilers now.  It's possible those are incorrect.  To be honest, I'm not sure how those settings were decided.  I suspect it's partly historical and they have not been reviewed for a while.  The main point of these CPPs is that they provide users with a flag to turn off these features if their compiler doesn't support them.  As a general rule, they should only be used when needed.  And we probably should review the current settings in the current machines to check whether we really do need them where they are currently set.

columnphysics/icepack_kinds.F90
```diff
@@ -17,10 +18,18 @@ module icepack_kinds
                             char_len_long  = 256, &
                             log_kind  = kind(.true.), &
                             int_kind  = selected_int_kind(6), &
+#if defined (NO_I8)
```
phil-blain, line 21: Minor point, but why `#if defined` and not `#ifdef` ? I think they do the same thing but maybe we should stick to one or the other?

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -8,7 +8,34 @@ Case Settings
...
+can be found in :ref:`cicecpps`.  The following CPPs are available.
+
+.. csv-table:: **CPP Macros**
+   :header: "CPP name", "description"
+   :widths: 15, 60
+
+   "",""
+   "**General Macros**", ""
+   "NO_I8", "Converts integer*8 to integer*4."
+   "NO_R16", "Converts real*16 to real*8."
```
phil-blain, line 33:
```suggestion
   "NO_I8", "Converts ``integer*8`` to ``integer*4``."
   "NO_R16", "Converts ``real*16`` to ``real*8``."
```

doc/source/user_guide/ug_running.rst
```diff
@@ -201,6 +201,25 @@ To add some optional settings, one might do::
 
 Once the cases are created, users are free to modify the **icepack.settings** and **icepack_in** namelist to further modify their setup.
 
+
+.. _cicecpps:
+
+C Preprocessing (CPP) Macros
```
phil-blain, line 207:
```suggestion
C Preprocessor (CPP) Macros
```

doc/source/user_guide/ug_running.rst
```diff
@@ -201,6 +201,25 @@ To add some optional settings, one might do::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are a few C Preprocessing Macros supported in the Icepack model.  These
```
phil-blain, line 210:
```suggestion
There are a few C Preprocessor Macros supported in the Icepack model.  These
```

doc/source/user_guide/ug_running.rst
```diff
@@ -201,6 +201,25 @@ To add some optional settings, one might do::
...
+support certain coding features to be excluded or included during the compile.  They
+exist in part to support the CICE model and other applications that use Icepack.
+
+For standalone Icepack, The CPPs are defined by the `CPPDEFS` variable in the Icepack
```
phil-blain, line 214:
```suggestion
For standalone Icepack, The CPPs are defined by the ``CPPDEFS`` variable in the Icepack
```

doc/source/user_guide/ug_running.rst
```diff
@@ -201,6 +201,25 @@ To add some optional settings, one might do::
...
+Makefile.  They are defined
+by passing the -D[CPP] to the C and Fortran compilers (ie. -DNO_I8) and this
+is what needs to be set in the `CPPDEFS` variable.  The value of `ICE_CPPDEFS` in
+**icepack.settings** is copied into the Makefile `CPPDEFS` variable as are settings
```
phil-blain, line 218:
```suggestion
is what needs to be set in the ``CPPDEFS`` variable.  The value of ``ICE_CPPDEFS`` in
**icepack.settings** is copied into the Makefile ``CPPDEFS`` variable as are settings
```

columnphysics/icepack_isotope.F90
```diff
@@ -482,10 +482,6 @@ function wiso_alpi(isp,tk)
 
       wiso_alpi = exp(alpai(isp)/tk**2 + alpbi(isp)/tk + alpci(isp))
 
-#ifdef NOFRAC
-      wiso_alpi = c1
-#endif
```
phil-blain, line 487: This is the only instance of "NOFRAC"... was this added for compiling CICE as part of another model ? (@eclare108213 you added that in 1ae0446 (Isotopes for Icepack (#307), 2020-04-06))

apcraig: According to @dabail10, this was added for debugging the isotopes and is not needed at this time.

columnphysics/icepack_kinds.F90
apcraig: In CICE, there is a mixture of #ifdef and #if defined.  The NO_I8 in CICE has #if defined, so I made the icepack format match that.  I could change it here, but then would probably want to change it in CICE as well.  If we want to take this step, I'm happy to do it.  I'm personally not that concerned about mixing the styles, although agree a single approach is nicer (but then requires diligence to maintain it).

PR #341, opened 2020-11-03
Add ustar_prev to while loop in icepack_atmo

proteanplanet:

This adds a necessary line for checking for convergence of the ustar loop in icepack_atmo that is missing.
Developer:  @proteanplanet 
Suggested Reviewers: @eclare108213 @dabail10 

This is a non-BFB change. There are no dependencies, and no new test cases are added.  Documentation is unchanged.

### DIFF
--- a/columnphysics/icepack_atmo.F90
+++ b/columnphysics/icepack_atmo.F90
@@ -271,6 +271,8 @@
 
       k = 1
       do while (abs(ustar - ustar_prev)/ustar > atmiter_conv .and. k <= natmiter)
+         k = k + 1
+         ustar_prev = ustar
 
          ! compute stability & evaluate all stability functions
          hol = vonkar * gravit * zlvl &
@@ -301,7 +303,6 @@
          tstar = rh * delt
          qstar = re * delq
 
-         k = k + 1
       enddo                     ! end iteration
 
       if (calc_strair) then


columnphysics/icepack_atmo.F90
```diff
@@ -271,6 +271,8 @@ psixh     ! stability function at zlvl   (heat and water)
 
       k = 1
       do while (abs(ustar - ustar_prev)/ustar > atmiter_conv .and. k <= natmiter)
+         k = k + 1
+         ustar_prev = ustar
```
dabail10, line 275: Looks like a bug fix! Thanks for catching this. Why are these two lines not at the end of the loop though?

proteanplanet: `ustar_prev = ustar` can't be at the end of the loop, because it would not be ustar_prev if it was.  It doesn't matter if `k=k+1` is at the start or end of the loop, but I have changed this so that it is consistent with the ocean coupling routine while loop now in CIME.

dabail10: You are correct! Looks good. Approving.

dabail10 (approved):

eclare108213 (approved):
Thanks for catching that, @proteanplanet

apcraig:
The code change seem to be mostly bit-for-bit but comprehensive test results can be seen here,

https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash#782a1b7d78ab0771d33aa411faeb8f06da657b92

PR #344, opened 2020-12-12
zsalinity test

eclare108213:
Adding smoke and restart tests for zsalinity in Icepack
- [x] Developer(s): 
@eclare108213 for this test, @njeffery for the original zsalinity code
    https://github.com/CICE-Consortium/Test-Results/wiki/3ca0dda63c.badger.intel.20-12-12.193908.0
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [x] Yes
    - [ ] No
    - [ ] Yes
        - [x] Yes?   There's a new -s zsal option
        - [ ] No 

This PR only adds and option and tests, with no changes to the original code.

@njeffery, please take a look at the output linked here for the smoke test, expecially the configuration.  The brine tracer seems to be working and the average salinity is changing, so the zsalinity code appears to be active.  I'm using the default forcing, and maybe this test would be better with one of the BGC forcing data sets (ISPOL or NICE)?  I have not yet been able to get a zsalinity case to run in CICE.
[logs.tar.zip](https://github.com/CICE-Consortium/Icepack/files/5683716/logs.tar.zip)

@apcraig @dabail10 Is the documentation for -s options entirely automated, or do I need to add something to readthedocs?

### DIFF
--- a/configuration/driver/icedrv_diagnostics_bgc.F90
+++ b/configuration/driver/icedrv_diagnostics_bgc.F90
@@ -794,7 +794,7 @@
       !-----------------------------------------------------------------
       ! NOTE these are computed for the last timestep only (not avg)
 
-         do n = 1, nx
+      do n = 1, nx
             pfzsal    = fzsal(n)
             pfzsal_g  = fzsal_g(n)
             phinS     = c0
@@ -850,7 +850,6 @@
             do k = 1,nilyr
                pSice(k) = trcr(n,nt_sice+k-1)
             enddo
-         enddo                  ! nx
 
       !-----------------------------------------------------------------
       ! start spewing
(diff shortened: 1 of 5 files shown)


njeffery:
Hi Elizabeth,


I would try running with z_tracers = true.  This shouldn't be needed, but it's possible that some  of the initialization/activation for zsalinity is tied to this flag.  If all is working well with zsalinity, then it shouldn't make a difference.


I'd stick with the default forcing.


n

apcraig:
For readthedocs to work, an update from master or an update to doc/requirements.txt will be needed.

eclare108213:
I'm not convinced that this test (or the zsalinity code) is working correctly.  In Icepack (this PR), the salinity changes slightly during a run but seems relatively impervious to major changes like different initial conditions, forcing, etc.  Its values are always like this:

avg salinity (ppt)     =        2.30068935292552013
avg salinity (ppt)     =        2.30068935292552057
avg salinity (ppt)     =        2.30068935292551879
avg salinity (ppt)     =        2.30068935292552146
avg salinity (ppt)     =        2.30068935292551924
avg salinity (ppt)     =        2.30068935292551924

In CICE (a draft PR I'll submit soon), the values look much more reasonable, although high.  (@njeffery turning z_tracers on did allow CICE to run, thanks.)

Since this needs to be understood, I don't think we should merge the PR as it is, but there are options:

1. just merge in the set_nml.zsal option for easier debugging purposes, since lots of namelist changes are required to get this to even run, but don't merge the changes to base_suite.
2. don't merge anything and just keep my branch around for debugging
3. merge all of it, with the possibility that the zsal test answers will change if/when debugged

I'm in favor of (1) but fine with closing this PR for now and going with (2).

apcraig:
I am fine with any of the 3 options.  I don't think it's a problem to add the test even if the answers change later if it's useful.  We are often changing results for a handful of tests at a time because of a change in implementation or otherwise.

eclare108213:
I fixed a bug in the initialization in which the solve_zsal parameter wasn't set after being read in from namelist, so it was reset to false when queried.  The diagnostics also weren't being printed correctly.  Now the zsalinity output from Icepack behaves similarly to CICE.  Note that z_tracers=T is not needed for Icepack, but is for CICE.

I'm comfortable with merging the zsalinity tests for both Icepack and CICE, understanding that there still may be bugs.  It's better to test...

apcraig (approved):

Issue #233, opened 2018-11-28
Bug in temperature_change_mushy


njeffery:
@eclare108213 This is a bugfix for a very infrequent failure in the mushy layer thermodynamics. The mushy temperature change iteration is currently set to a tolerance of ferrmax which is the same as the final energy conservation tolerance.  Every so often, when the temperature change iteration error is very close to ferrmax, the total energy conservation will fail because of round-off level errors in the thickness routine.  We corrected this in mpas_seaice by lowering the tolerance for the iteration to 0.9*ferrmax.  See below.  Lines with + indicated added changes and - indicates deletions.  This is a non BFB change.

We also corrected the diagnostics in the conservation check to properly indicate in this special case if the energy error is due to the temperature change or the thickness change.   This is BFB.

8 src/core_seaice/column/ice_therm_mushy.F90
@@ -523,7 +523,7 @@ subroutine two_stage_solver_snow(nilyr,       nslyr,      &
          ! check if solution is consistent 
          ! surface conductive heat flux should be less than 
          ! incoming surface heat flux
 -         if (fcondtop - fsurfn < ferrmax) then
+          if (fcondtop - fsurfn < 0.9_dbl_kind*ferrmax) then
@@ -576,7 +576,7 @@ subroutine two_stage_solver_snow(nilyr,       nslyr,      &
       ! check if solution is consistent 
       ! surface conductive heat flux should be less than 
       ! incoming surface heat flux
 -      if (fcondtop - fsurfn < ferrmax) then
+      if (fcondtop - fsurfn < 0.9_dbl_kind*ferrmax) then
@@ -842,7 +842,7 @@ subroutine two_stage_solver_nosnow(nilyr,       nslyr,      &
          ! check if solution is consistent 
          ! surface conductive heat flux should be less than 
          ! incoming surface heat flux
-          if (fcondtop - fsurfn < ferrmax) then
+         if (fcondtop - fsurfn < 0.9_dbl_kind*ferrmax) then
@@ -895,7 +895,7 @@ subroutine two_stage_solver_nosnow(nilyr,       nslyr,      &
       ! check if solution is consistent 
       ! surface conductive heat flux should be less than 
       ! incoming surface heat flux
 -      if (fcondtop - fsurfn < ferrmax) then
 +      if (fcondtop - fsurfn < 0.9_dbl_kind*ferrmax) then
15 src/core_seaice/column/ice_therm_vertical.F90
[...]

PR #352, opened 2021-03-15
modify error tolerance for rare mushy thermo failure

eclare108213:
This is a bugfix for a very infrequent failure in the mushy layer thermodynamics. The mushy temperature change iteration is currently set to a tolerance of ferrmax which is the same as the final energy conservation tolerance. Every so often, when the temperature change iteration error is very close to ferrmax, the total energy conservation will fail because of round-off level errors in the thickness routine. 
- [x] Developer(s): 
@njeffery @eclare108213 @dabail10 
Should not affect any of our tests -- this is expected to be BFB.
    - [x] bit for bit unless the results lie outside of the original error tolerance, which would trigger an abort.
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
This problem turned up in MPAS-seaice and was addressed there by lowering the error tolerance in ice_therm_mushy.F90 to 0.9*ferrmax, which will likely not be BFB.  Instead, we are loosening the final error tolerance comparison for this rare case, which will not affect any of our tests because they do not hit this problem (they would abort if they did).  This modification will allow the code to run without aborting in this special case.  See https://github.com/CICE-Consortium/Icepack/issues/233.

### DIFF
--- a/columnphysics/icepack_therm_vertical.F90
+++ b/columnphysics/icepack_therm_vertical.F90
@@ -1887,7 +1887,7 @@
            - fsnow*Lfresh - fadvocn) * dt
       ferr = abs(efinal-einit-einp) / dt
 
-      if (ferr > ferrmax) then
+      if (ferr > 1.1_dbl_kind*ferrmax) then
          call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
          call icepack_warnings_add(subname//" conservation_check_vthermo: Thermo energy conservation error" ) 
 


dabail10 (approved):

apcraig:
The icepack test suite is bit-for-bit with this PR.  Will test this in CICE next.

https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#8d68802687d01cf7224a2803be88394c8c3e5ced

apcraig:
The CICE test suite is bit-for-bit with this PR, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#714dba689bfd2ab53298c93206add1e90bf1de3a.

Issue #279, opened 2019-10-25
Negligible snow and convergence


eclare108213:
From @MichaelWinton at https://github.com/MichaelWinton/SIS2/blob/melt_pond/docs/icepack.md

With the GFDL pond scheme it turned out that negligible snow (ignored by Icepack) did not have negligible shortwave absorption when the energy conservation convergence criterion was applied in the iterative solver. A change was made to add snow shortwave absorption to the surface energy budget when the snow is so small as to be neglected.

eclare108213:
@proteanplanet @njeffery 
Just getting this issue back on our radar as something that we can consider fixing as part of the snow model upgrade for the MPAS/Icepack merge.

eclare108213:
@MichaelWinton 
I'd like to fix this in Icepack but I'm having trouble finding the change that you made - can't seem to make github compare your icepack_therm_mushy.F90 module with the Consortium's version (v1.0, right?).  Can you point me to the modified code, please, either as a code diff or just which lines in your code I should look at?
Thanks, e

dabail10:
I did add the shortwave "fix" to the mushy layer in CICE6, but only for the sea ice part. This is turned on via the sw_redist flag and is off by default for ktherm=2. The fraction moved is sw_frac and the temperature threshold is sw_dTemp.

Dave

    if (sw_redist) then

    dt_rhoi_hlyr = dt / (rhoi*hilyr)

    do k = 1, nilyr

       Iswabs_tmp = c0 ! all Iswabs is moved into fswsfc

       Tmlt = liquidus_temperature_mush(zSin(k))

       if (zTin(k) <= Tmlt - sw_dtemp) then
          ci = cp_ice - Lfresh * Tmlt / (zTin(k)**2)
          Iswabs_tmp = min(Iswabs(k), &
                           sw_frac*(Tmlt-zTin(k))*ci/dt_rhoi_hlyr)
       endif
       if (Iswabs_tmp < puny) Iswabs_tmp = c0

       dswabs = min(Iswabs(k) - Iswabs_tmp, fswint)

       fswsfc   = fswsfc + dswabs
       fswint   = fswint - dswabs
       Iswabs(k) = Iswabs_tmp

    enddo

    endif

eclare108213:
Here is how @MichaelWinton fixed the problem that turned up in SIS.  Is this fundamentally the same thing as yours, @dabail10 (except yours only occurs with sw_redist=T) ?  This seems reasonable to me.

`    else
[...]

PR #355, opened 2021-03-17
move shortwave absorbed in negligible snow layers to surface

eclare108213:
Bug fix to handle shortwave absorbed in very thin snow layers that are ignored by the thermo and shortwave routines.
- [x] Developer(s): 
    E. Hunke

Icepack base_suite including regression with current version:
148 measured results of 148 total results
148 of 148 tests PASSED
0 of 148 tests PENDING
0 of 148 tests MISSING data
0 of 148 tests FAILED

I'll try testing this in CICE next.

    - [x] bit for bit but I expect the results will change in lengthier or more varied tests
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
Addresses #279.  Instead of changing the shortwave fluxes only for the thermodynamic convergence criteria, as was done in #279, this modification moves the absorbed shortwave in thin snow layers to the surface, in the spirit of the redistributed shortwave changes in #326.  However the current redistribution is always done, not just when sw_redist=T, because it fixes a bug in the mushy thermo convergence criterion.
EDIT for clarification:  this change is always done for mushy thermo, but the problem may still exist for other thermodynamic options.  That's a science project.


eclare108213:
There's a problem with pushing test results to the wiki from my mac -- "invalid command code T" ?

> bash-3.2$ ./report_results.csh 
./report_results.csh: Running results.csh
Cloning into 'Test-Results.wiki'...
remote: Enumerating objects: 143, done.
remote: Counting objects: 100% (143/143), done.
remote: Compressing objects: 100% (71/71), done.
remote: Total 19620 (delta 117), reused 98 (delta 72), pack-reused 19477
Receiving objects: 100% (19620/19620), 16.80 MiB | 1013.00 KiB/s, done.
Resolving deltas: 100% (17019/17019), done.
./report_results.csh: writing to Test-Results.wiki/icepack_dev/e8df0ad846.conda.macos.21-03-17.201310.0.md
sed: 1: "Test-Results.wiki/icepa ...": invalid command code T
[master 4aefbf85] update e8df0ad8469b46b5ac6b32b4b42e52cd7f872a34 conda
 2 files changed, 54 insertions(+)
 create mode 100644 icepack_dev/e8df0ad846.conda.macos.21-03-17.201310.0.md
Enumerating objects: 8, done.
Counting objects: 100% (8/8), done.
Delta compression using up to 16 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 1.25 KiB | 1.25 MiB/s, done.
Total 5 (delta 4), reused 0 (delta 0)
remote: Resolving deltas: 100% (4/4), completed with 3 local objects.
To https://github.com/CICE-Consortium/Test-Results.wiki.git
   f227cd23..4aefbf85  master -> master

apcraig:
sed on mac is different than on linux.  we've had problems like this before with mac sed.  I'll try to look into it.  Ultimately, the script still seems to have worked, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#e8df0ad8469b46b5ac6b32b4b42e52cd7f872a34, so that's good at least.

eclare108213:
That's weird - it's on the icepack_by_hash_forks page but not on icepack_by_mach_forks page where I looked.

apcraig:
You're right, I hadn't noticed that.  That's a nice clue to where the problem is.

eclare108213:
CICE base-suite regression tests pass except for the usual 40-processor runs on badger.
https://github.com/CICE-Consortium/Test-Results/wiki/d627ea24f2.badger.intel.21-03-18.003159.0

I do expect this change to be non-BFB, but apparently the case 0 < hs < hs_min does not occur in our test suite.

I would like @dabail10 and @MichaelWinton to review this PR, since I've refactored the sw_redist logic and this fix is slightly different from what was done in SIS.  Thx

dabail10:
The more I think about this, this sounds like an inconsistency in hs_min between the radiation and the thermo.  Perhaps there could be a case where the snow is barely above hs_min during the radiation computation, but then it disappears before it gets to the thermo? @MichaelWinton, are you using the same hs_min for both?

eclare108213:
The radiation and the thermo are both keyed off of the hs_min parameter in our code (I'm not sure about SIS).  My assumption is that there can be Sswabs associated with nonzero hs < hs_min, which might be left over from changes in snow thickness such as the one you suggest, @dabail10.  Conservation of heat has never reached roundoff level in CICE, and probably it's at least partly due to leakages associated with these cutoff values not being properly dealt with - that's a bigger project to fix.

dabail10:
The problem with these type of errors is they are so intermittent. I'm not sure how to test for this. I think this solution is fine. Should we not do it for BL99 as well?

eclare108213:
I originally looked at doing it generally rather than just in mushy thermo, but that's where the sw_redist block was...  :).  I decided that decision could be made as part of a bigger project to fix the conservation issues.

dabail10 (approved):

MichaelWinton (approved):

apcraig:
Ready to merge?

PR #364, opened 2021-05-26
Add support for staggered atmospheric levels

phil-blain:
## PR checklist
    Add support for staggered atmospheric levels
- [X] Developer(s): 
    P. Blain based on code from F. Roy.
All tests are bit for bit with master: https://github.com/CICE-Consortium/Test-Results/wiki/5fcfe7d103.daley.intel.21-05-26.201401.0
    - [x] bit for bit
    - [ ] more substantial 
    - [X] Yes: https://github.com/CICE-Consortium/CICE/pull/603 depends on it.
    - [ ] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
In some applications, we might want the atmospheric input variables (wind, temperature and humidity) to be given at different vertical levels. For example, we might want to receive the winds at some level and the scalar variables (temperature and humidity) at another. This PR adds an optional argument `zlvs` to subroutine `atmo_boundary_layer` (and its callers), which corresponds to the vertical level at which the scalar variables are computed. The argument is optional so that so that higher-level drivers can optionally use the feature.

I also refactored the code a bit to make it more easily understandable.

--- 

This was implemented in our in-house versions of CICE4 and CICE5 and I'm porting it over to the Consortium as part of our project of updating our systems to CICE6.


eclare108213 (commented):
This looks okay to me, just reading through it.  Can you add a note about it in the text of the documentation, explaining how it works?

phil-blain:
@eclare108213 yes I'll do that.

phil-blain:
I've just pushed a new version of the last commit with some modifications to the Science guide. See the build here: https://cice-consortium-icepack--364.org.readthedocs.build/en/364/science_guide/sg_boundary_forcing.html#atmosphere

apcraig (commented):
There may be two places where an optional argument is passed into a subroutine without checking if that argument was passed into the calling subroutine.  This is one of the major issues with the current implementation of icepack.  My recommendation has been to create a local variable, set it to a default if the optional variable is not passed, then pass that local variable into other subroutines.  Something like

```
real (kind=dbl_kind) :: lzlvs

if (present(zlvs)) then
  lzlvs = zlvs
else
  lzlvs = zlvl
endif

call subroutine(..., zlvs=lzlvs, ....)

```
If there are better ways to handle this, I'd love to know!  I don't think the "optional" attribute is carried with the argument as you try to pass it down the calling tree, but maybe there is a way to do this?

phil-blain:
Yes, I thought about that myself when writing the code and checked my copy of _Modern Fortran Explained_. It is said that the `optional` attribute is carried forward though the call stack. That is why I did not use local variables as you suggest... 

Section 5.13 "Keyword and optional arguments" (p .89):
> In such a case, an absent optional argument is also regarded as absent in the second-level subprogram. [...] Such absent arguments may be propagated through any number of calls, provided the dummy argument is optional each case

apcraig:
Interesting.  I think we've run into problems before not checking optional arguments, but maybe I'm not remembering correct.  (See for example Qa_iso in icepack_atmo.F90).  If this is robust, that's going to be very helpful moving forward.  Maybe Qa_iso creates different problems because it's an array?

phil-blain:
Maybe some compiler implementations you were using are not conforming... might be worth fetching this branch and running the tests suite on more machines ? It builds and runs fine on Intel and Gfortran...

doc/source/science_guide/sg_boundary_forcing.rst
```diff
@@ -143,6 +144,10 @@ respectively, given the ice velocity :math:`\vec{U}_i`:
    \end{aligned}
    :label: stars
 
+
+Note that *atmo_boundary_layer* also accepts an optional argument, ``zlvs``, to support receiving scalar quantities from the atmospheric model (humidity and temperature)
+at a different vertical level than the winds. In that case a separate stability :math:`\Upsilon_s` is computed using the same formula as above but substituting :math:`z_o` by :math:`z_{o,s}`.
```
eclare108213, line 149: The "staggered" terminology makes sense here in the way that you use it, but it would be helpful to just add `(staggered)` after `at a different vertical level than the winds`, to more directly connect the comment below with this one.  Optional.

phil-blain: yes, good idea.

phil-blain: I've just pushed an updated commit that now reads 

> Note that *atmo_boundary_layer* also accepts an optional argument, ``zlvs``, to support staggered atmospheric levels, i.e. receiving scalar quantities from the atmospheric model (humidity and temperature)at a different vertical level than the winds.

eclare108213 (commented):
Was interfaces.include.orig committed intentionally?  Looks like it's a personal copy.

phil-blain:
Oups. I did `git add doc` so it was automatically included... it's not a personal copy, it's generated by the `gen_interfaces.sh` script and not cleaned up at the end...
https://github.com/CICE-Consortium/Icepack/blob/eb59c9524a99bdfc1351e182851d2deea8a7e3dd/doc/generate_interfaces.sh#L42

I'll re-push without  it.

eclare108213 (approved):

apcraig (approved):

Issue #149, opened 2018-02-10
Sensible heat


eclare108213:
In default runs without the SHEBA convergence data, the sensible heat shows small oscillations that need to be investigated.  Is this a problem with the thermodynamics solver?
![ice_diag full_itd_sensible_heat_flx](https://user-images.githubusercontent.com/8173983/36063920-708bbf34-0e40-11e8-9d5a-8e5acae2c2a0.png)

proteanplanet:
Hi Elizabeth,

May I ask some stupid questions about this, because it is of high interest to me.

I assuming that 'default runs' means no change to the namelist settings in icepack_parameters.F90, and includes the following settings:
-Bitz-Lipscomb
-highfreq is off
-natmiter=5

From this plot, one could be forgiven for thinking that some kind of filter has been used on the input SHEBA data between 0600Z observations. I assume we are looking at instantaneous values? 

All up, I agree, this is pretty odd.  I'm looking into it right now.

- Andrew

eclare108213:
We need to define better what we mean by "default".  In this case, it means the "out-of-the-box" run, i.e. with the namelist settings not altered.  This run used the default namelist settings except that it does NOT use the SHEBA convergence data.  So it's the mushy thermo, not BL99.  Another possibility is that the ridging scheme is doing something when it shouldn't be, although I would think that would show up in other fields besides the sensible heat flux.

Issue #361, opened 2021-05-18
Two questions about `atmo_boundary_layer`


phil-blain:
Hi everyone, I'm working on upstreaming some changes we've made to subroutine `atmo_boundary_layer` in module  **`icepack_atmo`**.

I have two questions about some part of the code that I'm not sure about:

1. In the iteration part of the subroutine, we compute the stability function in the stable case following Jordan et al 1999:
 https://github.com/CICE-Consortium/Icepack/blob/5cf223287d06167bb813d1fc2248258034511017/columnphysics/icepack_atmo.F90#L287-L294

However, at the end of the subroutine when we compute the diagnostic temperature `Tref` and humidity `Qref`, we instead use the  formulation  `\psi_m = -5 \zeta` (see for example [Kauffman & Large](https://github.com/CICE-Consortium/CICE/blob/master/doc/PDF/KL_NCAR2002.pdf) p.40). See the first line of the following code fragment:
https://github.com/CICE-Consortium/Icepack/blob/5cf223287d06167bb813d1fc2248258034511017/columnphysics/icepack_atmo.F90#L367-L374

 **Is this an implementation error ?** I'm guessing it should not make a big impact but for consistency shouldn't we use the same formulation in both places ?

---

2. At the end of the subroutine, when we compute the diagnostic temperature `Tref` and humidity `Qref`, we compute the stability parameter zeta (`hol` in the code) at height `zTrf`  by multiplying the existing `hol` value by `zTrf / zlvl`:
https://github.com/CICE-Consortium/Icepack/blob/5cf223287d06167bb813d1fc2248258034511017/columnphysics/icepack_atmo.F90#L364
instead of reusing the same formula as in the iteration (with `zTrf` instead of  `zlvl`):
https://github.com/CICE-Consortium/Icepack/blob/5cf223287d06167bb813d1fc2248258034511017/columnphysics/icepack_atmo.F90#L278-L281

Both approaches do not give the same value because the former implicitly uses `{u,q,t}star` from the penultimate iteration instead of the last iteration (which would be the case if we would use the longer formula)... 

**Is the underlying assumption that if we are converged it should not matter?** Because I've made a quick test and there can be some differences (in some cases it doubles the value of `hol`, others it's the same until the third decimal, sometimes it's at 1E-6, etc....) I'm not very familiar with the physics at play here so I don't know if in the end it would significantly change the model answers. I'm just sanity-checking what I'm understanding from my reading of the code...

[...]

PR #371, opened 2021-09-14
sensible+latent heatfluxes using linear bulk formula 

mhrib:
See Also: https://github.com/CICE-Consortium/CICE/pull/633

## PR checklist
    For stability boundary layer (atmbndy='default'): Use "traditional" linear bulk formula for sensible+latent heat fluxes but preserve stress formulation.
- [x] Developer(s): 
    @mhrib 
   @apcraig @eclare108213 
  77 measured results of 77 total results
  77 of 77 tests PASSED
  0 of 77 tests PENDING
  0 of 77 tests MISSING data
  0 of 77 tests FAILED
    - [x] bit for bit  (_for default settings_)
    - [x] more substantial 
    - [x] Yes  ( affects both CICE and Icepack)
    - [ ] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

By introducing the "simpler" linear bulk formula for sensible+latent heat fluxes, but preserve the temperature/humidity formulation for stress, the DMI-HYCOM-CICE (ERA5) code got more stable for long runs (several of years) without significant changes in the overall sea-ice cover/thickness. Without these modifications, the "Flux Error" (ferr) - rarely - became very high in special circumstances with low wind speeds and Tair>Tocn. In these situations the Jordan et al (1999) formulation might have it's limitations. These high Flux Errors seems to dissapear with the simpler formulations for sensible and latent heatfluxes for our setup at least.

The results are not BFB with the new setting turned on. However it is implemented turned off, which results in BFB results compared to the unmodified code.

NOTE:
* Both ICEPACK and CICE have to be updated, as the "atmbndy" namelist flag goes into both codes with default renamed to 'similarity'
* CICE PR: https://github.com/CICE-Consortium/CICE/pull/633


columnphysics/icepack_atmo.F90
```diff
@@ -359,8 +359,16 @@ subroutine atmo_boundary_layer (sfctype,            &
...
 
-      shcoef = rhoa * ustar * cp * rh + c1
-      lhcoef = rhoa * ustar * Lheat  * re
+      if (heatflux_linear) then
+         !- Use constant coefficients for sensible and latent heat fluxes
+         !    similar to atmo_boundary_const but using (wind-Vice) instead of wind-only
+         shcoef = (1.20e-3_dbl_kind)*cp_air*rhoa*vmag
+         lhcoef = (1.50e-3_dbl_kind)*Lheat *rhoa*vmag
+      else
+         !- Use Jordan et al (JGR, 1999) formulation
```
eclare108213, line 368: Please remove the ref to Jordan in the comment.  That's referring to a small modification of the overall turbulent heat flux calculation, which came from somewhere else (documented in readthedocs, I hope).

columnphysics/icepack_parameters.F90
```diff
@@ -250,11 +250,12 @@ module icepack_parameters
          TTTocn = 5107.4_dbl_kind      ! for qsat over ocn
 
       character (len=char_len), public :: &
-         atmbndy = 'default' ! atmo boundary method, 'default' ('ccsm3') or 'constant'
+         atmbndy = 'default'           ! atmo boundary method, 'default' ('ccsm3') or 'constant'
 
       logical (kind=log_kind), public :: &
          calc_strair     = .true.  , & ! if true, calculate wind stress
          formdrag        = .false. , & ! if true, calculate form drag
+         heatflux_linear = .false. , & ! if true, calculate sensible+latent heatfluxes using traditional linear bulk formula
```
eclare108213, line 258: As noted in the CICE PR, maybe there could be a third option for atmbndy instead of the new heatflux_linear flag.

eclare108213 (changes requested):
Let's discuss the namelist implementation.  A couple other comments.

There should also be changes to Icepack's documentation, in particular noting the additional option where sensible and latent heat are presented.  See sections 2.1.1 and 2.7.3.

eclare108213:
Does this fix https://github.com/CICE-Consortium/Icepack/issues/149?

columnphysics/icepack_parameters.F90
apcraig: If a namelist/argument exists that we can leverage to set this new option, that's preferable to creating a new namelist/argument when it makes sense.

mhrib:
https://github.com/CICE-Consortium/Icepack/pull/371#issuecomment-919604007

> Does this fix #149?

Not sure. But it is discussed further in #361 with figures pointing out the where the extreme values stems from.

columnphysics/icepack_parameters.F90
mhrib: Rewritten using 'atmbndy' namelist variable as suggested.
@eclare108213  I can't rename the PR removing 'traditional' but feel free to change it. Alternative I can create a new PR.

mhrib:
Update code using 'atmnbdy' which now allows three options: 'similarity' (default), 'constant', or 'mixed'
Also update documentation section 3.4.3.9.forcing_nml with the above info

eclare108213 (changes requested):
I like this much better, thanks.  My only request is to provide a little more information in the documentation.  In the “Atmosphere” section 2.1.1, add a note about the different options.  E.g. at the end of the third paragraph (add/fix formatting for readthedocs)

> … The resulting equations are provided here for the default boundary layer scheme, which is based on Monin-Obukhov theory (atmbndy = ‘stability’).  Alternatively, atmbndy = ‘constant’ provides constant coefficients for wind stress, sensible heat and latent heat calculations; atmbndy = ‘mixed’ uses the stability based calculation for wind stress and constant coefficients for sensible and latent heat fluxes.

mhrib:
Updated:
1) Update “Atmosphere” section 2.1.1 with almost the suggested text.
2) Abort if 'atmbndy' setting is unknown
3) For backward compability: 'atmbndy'=default is renamed to similarity

columnphysics/icepack_parameters.F90
```diff
@@ -250,7 +250,7 @@ module icepack_parameters
          TTTocn = 5107.4_dbl_kind      ! for qsat over ocn
 
       character (len=char_len), public :: &
-         atmbndy = 'default' ! atmo boundary method, 'default' ('ccsm3') or 'constant'
+         atmbndy = 'similarity'        ! atmo boundary method, 'similarity' ('ccsm3'), 'constant' or 'mixed'
```
phil-blain, line 253: I wonder if we should use this opportunity ("while we're here" kind of thing) to remove the `('ccsm3') `in the comment (and do the same in `icepack_init_parameters`.)  It has been there ever since the first commit (6f98a36) but 'ccsm3' was never a value for 'atmbndy', at least not in Icepack.


In fact it dates back to this commit: https://github.com/CICE-Consortium/CICE-svn-trunk/commit/dfa799dfb8a0eec78e4a976c82c25b869f3e959e and so it was never an acceptable value at all!

columnphysics/icepack_parameters.F90
```diff
@@ -646,12 +646,12 @@ subroutine icepack_init_parameters(   &
          TTTocn_in        ! for qsat over ocn
 
       character (len=*), intent(in), optional :: &
-         atmbndy_in ! atmo boundary method, 'default' ('ccsm3') or 'constant'
+         atmbndy_in   ! atmo boundary method, 'default' ('ccsm3') or 'constant'
```
phil-blain, line 649: Sorry to be picky, would it be possible to not introduce whitespace changes ? It adds noise to the diff (even though we can hide it in the GitHub UI or with `git diff` options, it is still an additional step to take).

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -314,8 +314,10 @@ forcing_nml
    :widths: 15, 15, 30, 15 
 
    "", "", "", ""
-   "``atmbndy``", "``constant``", "bulk transfer coefficients", "``default``"
-   "", "``default``", "stability-based boundary layer", ""
+   "``atmbndy``", "string", "bulk transfer coefficients", "``similarity``"
+   "", "``similarity``", "stability-based boundary layer", ""
```
phil-blain, line 318: Maybe here we could add something like
```suggestion
   "", "``similarity``", "stability-based boundary layer (previously: `default`)", ""
```

to help users that could be confused by their existing namelist having a now undocumented `default` value ?

phil-blain (commented):
Hi @mhrib, I think this looks good in general; I left a few comments / suggestions below.

phil-blain:
Sorry, I sent my review as you were pushing an updated version, so some of my feedback is already addressed :)

columnphysics/icepack_atmo.F90
```diff
@@ -942,6 +950,10 @@ subroutine icepack_atm_boundary(sfctype,                   &
 
       Cdn_atm_ratio_n = c1
 
+      if (trim(atmbndy) == 'default') then
+         atmbndy = 'similarity' ! For backward compability only ...
+      endif
+
```
phil-blain, line 956: It's a good idea to keep backwards compatibility. 
But I *think* that we now tend to do this kind of treatment directly in `ice_init` (in CICE, probably in Icepack the equivalent would be in `icedrv_init`)...

columnphysics/icepack_atmo.F90
```diff
@@ -971,6 +984,10 @@ subroutine icepack_atm_boundary(sfctype,                   &
                                    uvel=l_uvel, vvel=l_vvel,  &
                                    Uref=l_Uref, zlvs=zlvs)
          if (icepack_warnings_aborted(subname)) return
+      else
+         call icepack_warnings_add( &
+            subname//' atmbndy = '//trim(atmbndy)//' : unknown value')
+         call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
```
phil-blain, line 990: ...  and  the same thing here, maybe checking  for unknown values should also be in `icedrv_init` ? @apcraig @eclare108213 do we want to standardize on that for new code also in Icepack ?

doc/source/science_guide/sg_boundary_forcing.rst
```diff
@@ -120,7 +120,12 @@ parameterizations. Rain and all melted snow end up in the ocean.
 Wind stress and transfer coefficients for the
 turbulent heat fluxes are computed in subroutine
 *atmo\_boundary\_layer* following :cite:`Kauffman02`, with additions and changes as detailed in Appendix A of :cite:`Roberts15` for high frequency coupling (namelist variable ``highfreq``).
-The resulting equations are provided here.
+The resulting equations are provided here for the default boundary layer 
+scheme, which is based on Monin-Obukhov theory (``atmbndy`` = ‘stability’). 
+Alternatively, ``atmbndy`` = ‘constant’ provides constant coefficients for
+wind stress, sensible heat and latent heat calculations (computed in subroutine
+*atmo\_boundary\_const*); ``atmbndy`` = ‘mixed’ uses the stability based 
+calculation for wind stress and constant coefficients for sensible and latent heat fluxes.
```
phil-blain, line 128: Small formatting nit: I would avoid these "fancy" single quotes as copy pasting them directly to the namelist is asking for trouble :P Maybe just put the whole thing inside the double backticks, like this:


```suggestion
The resulting equations are provided here for the default boundary layer 
scheme, which is based on Monin-Obukhov theory (``atmbndy = 'stability'``). 
Alternatively, ``atmbndy = 'constant'`` provides constant coefficients for
wind stress, sensible heat and latent heat calculations (computed in subroutine
*atmo\_boundary\_const*); ``atmbndy = 'mixed'`` uses the stability based 
calculation for wind stress and constant coefficients for sensible and latent heat fluxes.
```

phil-blain (changes requested):

columnphysics/icepack_atmo.F90
mhrib: FYI. It is also done in CICE. So running CICE, you'll never ends up here - unless it is later changed in CICE.
If backward compatibility is only relevant for CICE, I'll suggest to remove this "default=similarity" from ICEPACK.

doc/source/user_guide/ug_case_settings.rst
mhrib: From CICE this is automatically changed from 'default' to 'similarity' together with a WARNING message printed to the log/stdout.

phil-blain (approved):

columnphysics/icepack_atmo.F90
apcraig: I raised this issue too.  We don't generally check "valid values" in Icepack.  There is quite a bit of that in CICE and the icepack driver.  However, users that incorporate Icepack in their own driver (like E3SM) will have to implement their own checking and won't be able to rely on Icepack to let them know if something is "off".  We might want to think about implementing explicit checks for Icepack arguments where it makes sense and is easy.  The would be a significant change but probably wouldn't change the model cost if done well (like checking values only when they are initialized/set or adding appropriate aborts in existing if logic).  But I think that is something we should discuss more broadly, create an issue if needed, and agree on general strategy before moving forward.

eclare108213: I agree, no need to do this check here, or even in this PR.  In my opinion it would be better to keep the original behavior (just the 'else' without values) so that 'similarity' or 'mixed' (or 'default' / anything else) would all work.  Long ago, I treated default conditions in the code as exactly that -- it didn't matter what the flag was set to -- as long as it wasn't something specific (like 'constant') then the code would execute the default behavior.  Setting the flag to 'default' in the namelist file only made that explicit.  We seem to be moving to a more rigorous implementation, which is fine.  I prefer to do namelist value checking upon initialization, for better efficiency during the timestepping.  As @apcraig says, it's better for now to plan to address this value-checking issue more generally and not worry about it here.  The easiest thing to do now is keep the original behavior.

eclare108213 (changes requested):

columnphysics/icepack_atmo.F90
mhrib: Updated without checking `atmbndy` valid setting:  'constant' or 'mixed'. Anything else end in 'similarity' behavior.

mhrib:
Updated without checking `atmbndy` valid setting: Choices: 'constant' or 'mixed'. Anything else end in 'similarity' behavior.

NOTE: In CICE, then file ice_init.F90 checks `atmbndy` for valid settings. This includes changing 'default' to 'similarity' for backward compatibility. This does also, somehow, follow the original treatment of 'default' settings as explained by @eclare108213 .

columnphysics/icepack_atmo.F90
```diff
@@ -359,8 +359,16 @@ subroutine atmo_boundary_layer (sfctype,            &
       ! as in Jordan et al (JGR, 1999)
       !------------------------------------------------------------
 
-      shcoef = rhoa * ustar * cp * rh + c1
-      lhcoef = rhoa * ustar * Lheat  * re
+      if (trim(atmbndy) == 'mixed') then
+         !- Use constant coefficients for sensible and latent heat fluxes
+         !    similar to atmo_boundary_const but using vmag instead of wind
+         shcoef = (1.20e-3_dbl_kind)*cp_air*rhoa*vmag
+         lhcoef = (1.50e-3_dbl_kind)*Lheat *rhoa*vmag
```
proteanplanet, line 366: Please remove literal constants from the code (here 1.20e-3_dbl_kind  and 1.50e-3_dbl_kind) and place them in the Icepack constants file.  Also, here, how does this work if `sfctype(1:3)=='ocn'`?

proteanplanet (changes requested):

configuration/scripts/options/set_nml.dyn
```diff
@@ -3,5 +3,5 @@ kstrength = 0
 krdg_partic = 0
 krdg_redist = 0
 formdrag = .false.
-atmbndy = 'constant'
+atmbndy = 'similarity'
```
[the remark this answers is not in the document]
mhrib: This was an error from my site. Will revert atmbndy to constant

columnphysics/icepack_atmo.F90
mhrib: Introduce p0012, p0015, which makes sense.

columnphysics/icepack_atmo.F90
mhrib: Done...

eclare108213 (approved):

columnphysics/icepack_atmo.F90
```diff
@@ -503,8 +512,8 @@ subroutine atmo_boundary_const (sfctype,  calc_strair, &
       ! coefficients for turbulent flux calculation
       !------------------------------------------------------------
 
-      shcoef = (1.20e-3_dbl_kind)*cp_air*rhoa*wind
-      lhcoef = (1.50e-3_dbl_kind)*Lheat *rhoa*wind
+      shcoef = p0012*cp_air*rhoa*wind
+      lhcoef = p0015*Lheat *rhoa*wind
```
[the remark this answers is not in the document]
mhrib: Good point, which I did not really catch.
Changed "general constants" p0012/p0015 to "specific constants" senscoef/latncoef.
  (p0012/p0015 not used elsewhere)

eclare108213 (approved):

columnphysics/icepack_atmo.F90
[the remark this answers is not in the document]
proteanplanet: @eclare108213 @mhrib Yes, this is fine, and what I had in mind.

Issue #390, opened 2022-05-03
update_ocn_f problem


dabail10:
Not sure when this code came in, but there is a bug here. When update_ocn_f is true, we do need to compute the freshwater and salt due to frazil formation even for ktherm == 2. This was supposed to be "somewhere" else, but I don't see where it is done.

```
 if (update_ocn_f) then
         if (ktherm <= 1) then
            dfresh = -rhoi*vi0new/dt 
            dfsalt = ice_ref_salinity*p001*dfresh
            fresh  = fresh + dfresh
            fsalt  = fsalt + dfsalt
         ! elseif (ktherm == 2) the fluxes are added elsewhere
         endif
      else ! update_ocn_f = false
         if (ktherm == 2) then ! return mushy-layer frazil to ocean (POP)
            vi0tmp = fnew*dt / (rhoi*Lfresh)
            dfresh = -rhoi*(vi0new - vi0tmp)/dt
            dfsalt = ice_ref_salinity*p001*dfresh
            fresh  = fresh + dfresh
            fsalt  = fsalt + dfsalt
            frazil_diag = frazil - vi0tmp
         ! elseif ktherm==1 do nothing
         endif
      endif
```

eclare108213:
I think it's in ice_comp_mct.F, at least in E3SM.  The logic associated with update_ocn_f is extremely convoluted and confusing.  I'd like to fix it as part of the Icepack merge effort - will need to discuss how to implement it broadly, since it likely will affect many/most coupled model configurations.

See lines 1973ff and 2555-2556 in
https://github.com/E3SM-Project/E3SM/blob/302618e817d9c4299a213b62b49707fcf2c5b01e/components/mpas-seaice/driver/ice_comp_mct.F.

These changes were made at the same time as removing the temperature limiting mentioned in
https://github.com/CICE-Consortium/Icepack/issues/385.

dabail10:
What was the solution in E3SM here? I had to remove the if block in the update_ocn_f = .true. code. This was done in #418 . So, it is answer changing as the "bug" here was the the fresh water and salt flux were not computed for frazil formation here. This is done somewhere else in E3SM.

eclare108213:
[...]

PR #392, opened 2022-06-16
Add salt flux computation option

dabail10:

## PR checklist
This adds the option to use the prognostic salinity in the salt flux computation.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#7c33d8ebeaa42d3d76767b7825730a9c2a555ec2
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
This adds saltflux_option to the namelist for icepack. The options are '4psu' or 'prognostic'. I also removed the if (ktherm == 1) condition around the code for update_ocn_f in ice_therm_itd.F90 (add_new_ice). We need to compute these fluxes regardless of the ktherm option. I'm not sure where in MPAS-SeaIce the salt and fresh water fluxes are computed. Can we use update_ocn_f here? Related to issue #390 Added test results comparing to cheyenne baselines.

### DIFF
--- a/columnphysics/icepack_shortwave.F90
+++ b/columnphysics/icepack_shortwave.F90
@@ -2057,8 +2057,8 @@
          gaer                     , & ! total aerosol asymmetry parameter
          swdr                     , & ! shortwave down at surface, direct  (W/m^2)
          swdf                     , & ! shortwave down at surface, diffuse (W/m^2)
-         rnilyr                   , & ! real(nilyr)
-         rnslyr                   , & ! real(nslyr)
+         rnilyr                   , & ! 1. / real(nilyr)
+         rnslyr                   , & ! 1. / real(nslyr)
          rns                      , & ! real(ns)
          tmp_0, tmp_ks, tmp_kl        ! temp variables
 
@@ -2249,7 +2249,7 @@
          fm_pnd = 0.50_dbl_kind    ! ponded ice fraction of scat coeff for - stn dev in alb
 
       real (kind=dbl_kind),  parameter :: &   !chla-specific absorption coefficient
-         kchl_tab = 0.01 !0.0023-0.0029 Perovich 1993, also 0.0067 m^2 (mg Chl)^-1
+         kchl_tab = p01  !0.0023-0.0029 Perovich 1993, also 0.0067 m^2 (mg Chl)^-1
                          ! found values of 0.006 to 0.023 m^2/ mg  (676 nm)  Neukermans 2014
                          ! and averages over the 300-700nm of 0.0075 m^2/mg in ice Fritsen (2011)
                          ! at 440nm values as high as 0.2 m^2/mg in under ice bloom (Balch 2014)
@@ -2443,13 +2443,13 @@
 
                  ! get grain size index:
                  ! works for 25 < snw_rds < 1625 um:
-                 if (tmp_gs < 125) then
-                   tmp1 = tmp_gs/50
+                 if (tmp_gs < 125._dbl_kind) then
+                   tmp1 = tmp_gs/(50._dbl_kind)
                    k_bcini(k) = nint(tmp1)
-                 elseif (tmp_gs < 175) then
+                 elseif (tmp_gs < 175._dbl_kind) then
                    k_bcini(k) = 2
                  else
-                   tmp1 = (tmp_gs/250)+2
+                   tmp1 = (tmp_gs/250._dbl_kind)+c2
                    k_bcini(k) = nint(tmp1)
                  endif
               else                  ! use the largest snow grain size for ice
@@ -2464,10 +2464,10 @@
               ! check bounds:
               if (k_bcini(k) < 1)  k_bcini(k) = 1
               if (k_bcini(k) > 8)  k_bcini(k) = 8
-              if (k_bcins(k) < 1)  k_bcins(k) = 1
-              if (k_bcins(k) > 10) k_bcins(k) = 10
-              if (k_bcexs(k) < 1)  k_bcexs(k) = 1
-              if (k_bcexs(k) > 10) k_bcexs(k) = 10
+!              if (k_bcins(k) < 1)  k_bcins(k) = 1
+!              if (k_bcins(k) > 10) k_bcins(k) = 10
+!              if (k_bcexs(k) < 1)  k_bcexs(k) = 1
+!              if (k_bcexs(k) > 10) k_bcexs(k) = 10
 
               ! print ice radius index:
               ! write(warnstr,*) subname, "MGFICE2:k, ice index= ",k,  k_bcini(k)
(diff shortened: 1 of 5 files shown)


njeffery:
@dabail10 :  Should saltflux_option be added to Icepack_in at this point?

dabail10:
Yes! Good catch.

dabail10:
Technically, it should also be added to the Icepack driver as well.

njeffery:
Okay, I'll wait for that.

Also, do any of the Icepack testcases  verify salt conservation?

dabail10:
Unfortunately, there are no conservation diagnostics in Icepack. Something on my to-do list.

dabail10:
We do check salt conservation in CICE.

dabail10:
Good question. I've never turned that on.

eclare108213:
I think conserv_check is for checking algorithms like incremental remapping or ridging -- it used to be hardwired in the code for debugging during development, and was later moved out to the namelist for easier automated testing.

columnphysics/icepack_itd.F90
```diff
@@ -1953,8 +1970,8 @@ subroutine icepack_init_itd_hist (ncat, hin_max, c_hi_range)
             write (c_nc, '(i2)') n    
 
             ! Write hin_max to character string
-            write (c_hinmax1, '(f7.3)') hin_max(n-1)
-            write (c_hinmax2, '(f7.3)') hin_max(n)
+            write (c_hinmax1, '(f6.3)') hin_max(n-1)
+            write (c_hinmax2, '(f6.3)') hin_max(n)
```
apcraig, line 1974: What was the format here changed?

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -359,7 +359,8 @@ forcing_nml
    "", "``mm_per_sec``", "(same as MKS units)", ""
    "", "``m_per_sec``", "", ""
    "``restore_ocn``", "logical", "restore sst to data", "``.false.``"
-   "``tfrz_option``", "``linear_salt``", "linear functino of salinity (ktherm=1)", "``mushy``"
+   "``saltflux_option``", "``4psu``","``prognostic``"
```
apcraig, line 362: The syntax here is not correct.

apcraig:
Just want to confirm everything is bit-for-bit in Icepack and CICE?  Do you have a link to any test results?

doc/source/user_guide/ug_case_settings.rst
dabail10: Good catch.

njeffery:
@dabail10 :  Removing ktherm <=1 for update_ocn_f = true would be a problem for the MPAS-seaice implementation. This is because the freshwater and salt fluxes from the freeze melt potential are computed originally by the ocean model and corrected to be consistent with mushy-layer in the driver, ice_comp_mct.  If the correction is done in add_new_ice instead of the driver then it should look like the update_ocn_f = false version dfresh = -rhoi*(vi0new-vi0tmp)/dt , not dfresh = -rhoi*vi0new/dt.

njeffery:
@dabail10 : Do you know of any other code that's impacted by update_ocn_f?  Right now E3SM runs with update_ocn_f = true and switching to false would break mass/salt conservation without some code change.

dabail10:
I guess I am not understanding correctly. I'm not sure about other coupled models using update_ocn_f, maybe NOAA ... my understanding is that update_ocn_f = .true. means that the sea ice component computes the fluxes of freshwater and salt due to frazil formation. The ocean component computes it when update_ocn_f = .false. and then the sea ice component sends a correction for the "extra" frazil. The rationale for this is that the sea ice component would compute the freshwater and salt *and* use the ktherm == 2 actual salinity.

columnphysics/icepack_therm_vertical.F90
```diff
@@ -241,7 +242,7 @@ subroutine thermo_vertical (nilyr,       nslyr,     &
          einter          ! intermediate energy
 
       real (kind=dbl_kind) :: &
-         fadvocn ! advective heat flux to ocean
+         fadvocn, saltvol, dfsalt ! advective heat flux to ocean
```
eclare108213, line 245: Put the new variables on separate lines so that they can be identified with ! comments (add units too please)

columnphysics/icepack_parameters.F90
```diff
@@ -162,6 +162,11 @@ module icepack_parameters
                                   ! 'linear_salt' = -depressT * sss
                                   ! 'mushy' conforms with ktherm=2
 
+      character(len=char_len), public :: &
+         saltflux_option  = '4psu'! Salt flux computation
```
eclare108213, line 166: It would be better to add a separate parameter for the reference value itself, rather than hardwiring it for 4 psu.  What if some other model prefers 4.5 psu?  You could use prognostic and constant as the types for now, and then maybe someone would add some other type in the future (seasonal? stochastic?).

Also, "saltflux_option" sounds clunky to me, maybe because it's not immediately clear what kind of "option" it is.  We use "type" in other flags, e.g. bgc_flux_type.  The equivalent here would work:  salt_flux_type.  (And yes I realize that there's a tfrz_option, so this revision is optional...)

eclare108213 (commented):
The logic associated with update_ocn_f, ktherm, and now saltflux_option is really convoluted.  Adding this new option only makes it worse.  Can we work out what the logic ought to be and fix this?  

Mushy handles the flux calculations quite differently from the other thermodynamics options.  Can the original flags be pushed down into the mushy routines and take care of things there?  

It seems that saltflux_option='prognostic' only when ktherm=2 (mushy), but ktherm=2 does not imply that saltflux_option='prognostic' (it can be constant, e.g. 4 psu).  So there are three separate issues, partly related and all interacting:  1) whether the ocean assumes that frazil ice is included in the water and salt fluxes, controlled by update_ocn_f, 2) whether the salinity coupling is via an assumed constant value or calculated prognostically by the thermodynamics, controlled by saltflux_option, and 3) whether the ice salinity is prognostic or not, controlled by ktherm.  There needs to be a check (in CICE too!) that if saltflux_option='prognostic' then ktherm=2, otherwise abort.

Am I understanding this correctly?

dabail10:
Thanks for your feedback @eclare108213 ! I was following the example of tfrz_option as you mention. Here we do minus1p8 instead of constant. I agree that it makes more sense to have a constant option for both and let the user assign the value. The issue with saltflux_option is that it is needed in the frazil ice (add_new_ice) in icepack_therm_itd.F90. It is also needed in the zap routines. The way I have it set up now is that update_ocn_f and saltflux_option are independent. You could imagine someone wants to actually do a prognostic salinity flux whether or not the ocean component is doing the freshwater and salt for frazil or not. Also, I would imagine that you could even want a prognostic salinity flux with ktherm == 1 or maybe a different constant value. I thought about doing something like only computing the prognostic salt flux when ktherm == 2.

PR #400, opened 2022-09-20
dEdd aerosol fixes

apcraig:
## PR checklist
    dEdd aerosol fixes
- [X] Developer(s): 
    apcraig, eclare108213
    bit-for-bit against current Icepack trunk, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#0717620693e6545c0c693cb17987a5e412162b79.  Also testing CICE suite on cheyenne.
    - [X] bit for bit except when tr_aero is on (added to test suite)
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes, modal
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- update g/w0/tau calculation, get rid of use of puny
- add missing g/w0/tau calculation after "aerosol in snow" calculation
- fix bug in computation of modal_aero taer/waer/gaer, changed divide to multiple of rnilyr/rnslyr
- refactor divides of rnilyr/rnslyr to multiplies of real(nilyr/nslyr)
- add modal option, update test suite

### DIFF
--- a/columnphysics/icepack_shortwave.F90
+++ b/columnphysics/icepack_shortwave.F90
@@ -2249,7 +2249,7 @@
          fm_pnd = 0.50_dbl_kind    ! ponded ice fraction of scat coeff for - stn dev in alb
 
       real (kind=dbl_kind),  parameter :: &   !chla-specific absorption coefficient
-         kchl_tab = 0.01 !0.0023-0.0029 Perovich 1993, also 0.0067 m^2 (mg Chl)^-1
+         kchl_tab = p01  !0.0023-0.0029 Perovich 1993, also 0.0067 m^2 (mg Chl)^-1
                          ! found values of 0.006 to 0.023 m^2/ mg  (676 nm)  Neukermans 2014
                          ! and averages over the 300-700nm of 0.0075 m^2/mg in ice Fritsen (2011)
                          ! at 440nm values as high as 0.2 m^2/mg in under ice bloom (Balch 2014)
@@ -2443,13 +2443,13 @@
 
                  ! get grain size index:
                  ! works for 25 < snw_rds < 1625 um:
-                 if (tmp_gs < 125) then
-                   tmp1 = tmp_gs/50
+                 if (tmp_gs < 125._dbl_kind) then
+                   tmp1 = tmp_gs/(50._dbl_kind)
                    k_bcini(k) = nint(tmp1)
-                 elseif (tmp_gs < 175) then
+                 elseif (tmp_gs < 175._dbl_kind) then
                    k_bcini(k) = 2
                  else
-                   tmp1 = (tmp_gs/250)+2
+                   tmp1 = (tmp_gs/250._dbl_kind)+c2
                    k_bcini(k) = nint(tmp1)
                  endif
               else                  ! use the largest snow grain size for ice
@@ -2464,10 +2464,10 @@
               ! check bounds:
               if (k_bcini(k) < 1)  k_bcini(k) = 1
               if (k_bcini(k) > 8)  k_bcini(k) = 8
-              if (k_bcins(k) < 1)  k_bcins(k) = 1
-              if (k_bcins(k) > 10) k_bcins(k) = 10
-              if (k_bcexs(k) < 1)  k_bcexs(k) = 1
-              if (k_bcexs(k) > 10) k_bcexs(k) = 10
+!              if (k_bcins(k) < 1)  k_bcins(k) = 1
+!              if (k_bcins(k) > 10) k_bcins(k) = 10
+!              if (k_bcexs(k) < 1)  k_bcexs(k) = 1
+!              if (k_bcexs(k) > 10) k_bcexs(k) = 10
 
               ! print ice radius index:
               ! write(warnstr,*) subname, "MGFICE2:k, ice index= ",k,  k_bcini(k)
@@ -3158,10 +3158,8 @@
 
          ! bgc layer
          fswpenl(k) = fswpenl(k) + fthrul(k)* fi
-         if (k == nilyr) then
-            fswpenl(k+1) = fswpenl(k+1) + fthrul(k+1)*fi
-         endif
       enddo                     ! k
+      fswpenl(nilyr+1) = fswpenl(nilyr+1) + fthrul(nilyr+1)*fi
 
 #ifdef UNDEPRECATE_0LAYER
       !----------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


dabail10:
I have to test this out in CESM.

Dave

columnphysics/icepack_shortwave.F90
```diff
@@ -2710,20 +2711,18 @@ subroutine compute_dEdd (nilyr,    nslyr,    klev,  klevp,  &
...
+                                      (tau(k) + taer)
                   tau(k) = tau(k) + taer
                enddo       ! k
             endif     ! tr_aero
 
             ! pond
          else !if( srftyp == 2 ) then
             ! pond water layers evenly spaced
-            dz = hp/(c1/rnslyr+c1)
+            dz = hp/(real(nslyr,kind=dbl_kind)+c1)
```
eclare108213, line 2725: I think this change is correct, but I question the original expression -- does anyone understand why the +1 is in the denominator?  It looks like the dEdd scheme is dividing  the pond depth (volume) evenly among the snow layers, but then why is this not `dz=hp/nslyr` ?

eclare108213 (approved):
This all looks correct to me.  Since standard testing is BFB, the compiler/optimizer must be fixing the 1/(1/n)=n coding issues. Testing with tr_aero=T in a realistic simulation would be helpful, as a sanity check -- I would not expect those results to be BFB -- although these bugs should be fixed, regardless. Thanks much.

columnphysics/icepack_shortwave.F90
dabail10: I believe this was the "surface scattering layer" versus the "internal layer". So, I think this is fine.

eclare108213: Okay, now I see that the loop immediately following this line is over k = 0, nslyr.  So the ponds are simply divided into nslyr+1 equally-spaced layers for the dEdd calculations, and I won't worry about what "SSL" might mean for a pond -- it's just bookkeeping for consistency with the snow calculation, I guess.  Thanks.

columnphysics/icepack_shortwave.F90
```diff
@@ -2654,8 +2652,11 @@ subroutine compute_dEdd (nilyr,    nslyr,    klev,  klevp,  &
                endif  !modal_aero
 !mgf--
                enddo       ! na
-               gaer = gaer/(waer+puny)
-               waer = waer/(taer+puny)
+               g(k)   = (g(k)*w0(k)*tau(k) + gaer) / &
+                             (w0(k)*tau(k) + waer)
+               w0(k)  =      (w0(k)*tau(k) + waer) / &
+                                   (tau(k) + taer)
+               tau(k) = tau(k) + taer
```
dabail10, line 2659: It does look like these three lines will impact tr_aero and model_aero = .false. I looked back in the CICE5 code and this was the original implementation. Trying it out now.

columnphysics/icepack_shortwave.F90
```diff
@@ -2667,34 +2668,34 @@ subroutine compute_dEdd (nilyr,    nslyr,    klev,  klevp,  &
...
                       taer = taer + &
-                           (aero_mp(na+1)/rnslyr)*kaer_tab(ns,(1+(na-1)/4))
+                           (aero_mp(na+1)*rnslyr)*kaer_tab(ns,(1+(na-1)/4))
                       waer = waer + &
-                           (aero_mp(na+1)/rnslyr)*kaer_tab(ns,(1+(na-1)/4))* &
+                           (aero_mp(na+1)*rnslyr)*kaer_tab(ns,(1+(na-1)/4))* &
                            waer_tab(ns,(1+(na-1)/4))
                       gaer = gaer + &
-                           (aero_mp(na+1)/rnslyr)*kaer_tab(ns,(1+(na-1)/4))* &
+                           (aero_mp(na+1)*rnslyr)*kaer_tab(ns,(1+(na-1)/4))* &
```
dabail10, line 2698: This is almost comical ...

columnphysics/icepack_shortwave.F90
```diff
@@ -2710,20 +2711,18 @@ subroutine compute_dEdd (nilyr,    nslyr,    klev,  klevp,  &
...
-                  gaer = gaer/(waer+puny)
-                  waer = waer/(taer+puny)
-                  g(k)   = (g(k)*w0(k)*tau(k) + gaer*waer*taer) / &
-                           (w0(k)*tau(k) + waer*taer)
-                  w0(k)  = (w0(k)*tau(k) + waer*taer) / &
-                           (tau(k) + taer)
+                  g(k)   = (g(k)*w0(k)*tau(k) + gaer) / &
+                                (w0(k)*tau(k) + waer)
+                  w0(k)  =      (w0(k)*tau(k) + waer) / &
+                                      (tau(k) + taer)
```
dabail10, line 2717: This also looks like it might impact answers.

columnphysics/icepack_shortwave.F90
apcraig: The changes to the "puny" implementation elsewhere will change answers but only slightly.  On the other hand, this block of changes was a bug and could be more consequential.

columnphysics/icepack_shortwave.F90
apcraig: I guess this is just modal aerosol.  Not sure if you or others are using that.  We haven't been testing it in CICE because it hasn't been working.  But now it does and I have added tests to both Icepack and CICE.  Still can't say whether answers are scientifically correct though.

columnphysics/icepack_shortwave.F90
apcraig: These changes should be round-off-ish.

dabail10:
After four years, the answers are definitely different. I want to run it longer to see if it is "climate" changing.

eclare108213:
I wouldn't be surprised if these mods are "climate changing" when tr_aero=T and modal_aero=F, when comparing CICE6 before/after this PR.  Is that what you're testing @dabail10?  

These are bugs in CICE6 that were not present in CICE5, and should be fixed regardless of whether they are deemed to be climate changing, right away.  If this modified code produces reasonable-looking answers, then I think we should merge it and keep moving forward.  More careful comparisons with CICE5 might be needed to understand differences between the code versions, but let's at least get the obvious bugs fixed ASAP.  We may need to make further modifications later, as we continue the BGC merge.  Is this plan okay with those of you reading this thread?

apcraig:
@eclare108213, I agree that these things need to be fixed.  We should try to understand the impact of the bug and document that, but that could be done over days/weeks as needed.

eclare108213:
@dabail10 How are your runs going? I'd like to merge this one.

dabail10:
I have 20 years of both. Here are some diagnostics:

https://webext.cgd.ucar.edu/B1850/b.cesm3_cam058_mom_e.B1850WscMOM.ne30_L58_t061.camdev_cice6.026b2/ice/b.cesm3_cam058_mom_e.B1850WscMOM.ne30_L58_t061.camdev_cice6bugfix.026b-b.cesm3_cam058_mom_e.B1850WscMOM.ne30_L58_t061.camdev_cice6.026b/yrs1-21/

Basically I can't really tell as it is only 20 years, but it does not appear to be climate changing. I will run these out for 100 years.

Dave

apcraig:
Can we merge while we continue to assess?

dabail10:
I would say go ahead.

Issue #401, opened 2022-09-28
closing/flushing nu_diag at end


cmbitz:
When I compiled with netcdf, I had to add a
close(nu_diag) 
or
flush(nu_diag)
at the end of the program icedrv, otherwise I don't get the last output written to the log file, including the ICEPACK COMPLETED statement, and the stdio would echo "ICEPACK run did not complete".

I tried making it a
call icepack_warnings_flush(nu_diag)
but it didn't work

I'm using my macbook pro with Monterey OSX and gnu compiler. I didn't have this problem until I compiled with netcdf (didn't matter whether I wrote the history file though). With this change all is well getting it to run. 
Thanks, Cecilia

eclare108213:
HI CC,
I also have been having this problem and hadn't fixed it.  Which OSX are you running?  I'm having trouble with Monterey 12.4 on my regular laptop, using the conda environment.  The loaner laptop last week did not have this issue, running OSX Big Sur 11.6.2.  If the close statement is enough to fix it, let's do that...

cmbitz:
I'm running OSX Monterey 12.6. Thanks for entertaining the fix for me, though it is the OS's problem.

apcraig:
I hope this is fixed.  We are flushing and closing all the fileunits at the end of the run.  I am going to close this issue, but please reopen if the problems persist.

PR #403, opened 2022-10-06
close nu_diag

eclare108213:
    Close the diagnostics file at the end of the program. 
- [x] Developer(s): 
    @cmbitz @eclare108213 
    This change compiles, runs, and completes properly on my laptop.  Results from base_suite without regression:
> 131 measured results of 131 total results
> 131 of 131 tests PASSED
> 0 of 131 tests PENDING
> 0 of 131 tests MISSING data
> 0 of 131 tests FAILED

    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

Addresses #401. Without closing the file, "ICEPACK COMPLETED" was not written/flushed to the diagnostics file in a newer Mac OS and the scripts returned a "not completed" message.

### DIFF
--- a/configuration/driver/icedrv_InitMod.F90
+++ b/configuration/driver/icedrv_InitMod.F90
@@ -163,6 +163,7 @@
       subroutine init_restart
 
       use icedrv_calendar, only: time, calendar
+      use icedrv_constants, only: nu_restart
       use icepack_intfc, only: icepack_aggregate
       use icedrv_domain_size, only: ncat, max_ntrcr, nx
       use icedrv_init, only: ice_ic
@@ -216,6 +217,8 @@
          call init_bgc
          if (restart) call read_restart_bgc ! complete BGC initialization
       endif
+
+      close (nu_restart)
 
       !-----------------------------------------------------------------
       ! aggregate tracers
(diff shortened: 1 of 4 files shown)


apcraig:
Recommend adding

```
! Should not be needed, bug in some MacOS implementations
close(ice_stdout)
close(nu_diag)
do n = 1,nx
   close (nu_diag_out+n-1)
enddo
```

apcraig:
Also, I'm not sure I'd call this a bug.  This is a workaround for a system bug outside Icepack :)

apcraig (changes requested):
Suggest extending the fix a bit more, see prior comment.

eclare108213:
> Also, I'm not sure I'd call this a bug. This is a workaround for a system bug outside Icepack :)

Yes, but isn't it best practice to close all open files before ending a program, to avoid these kinds of problems?  Maybe that's a 30-year-old best practice that's less relevant today, until suddenly it's relevant again...

I'll extend as suggested.  Thanks!

eclare108213:
I also closed the restart file and the SHEBA forcing file; now running a base_suite to make sure everything compiles, runs and completes.  Regression is difficult because the original code doesn't complete on my laptop, but that could be tested on a different machine.

eclare108213:
Base_suite now completes - summary results are posted in the PR template above.

apcraig (approved):

eclare108213:
I believe there is still a problem with the grid-cell diagnostic files not being completely written before the program ends, so keeping this issue open for now.

PR #415, opened 2022-11-25
bug fixes for snow grain radius, brine conservation check

eclare108213:
This PR corrects the magnitude of a parameter controlling wet metamorphosis of snow (answer changing), and fixes a brine volume conservation check (does not affect the solution).
- [x] Developer(s): 
@eclare108213 

160 measured results of 160 total results
156 of 160 tests PASSED
0 of 160 tests PENDING
2 of 160 tests MISSING data
2 of 160 tests FAILED

Both runs with -s snwgrain have different answers, as expected:
FAIL conda_macos_smoke_col_1x1_debug_run1year_snw30percent_snwgrain compare ibased33 different-data
FAIL conda_macos_restart_col_1x1_snwgrain_snwitdrdg compare ibased33 different-data

This was a change in a recent PR, after the baseline was run - not a problem here:
MISS conda_macos_smoke_col_1x1_debug_fsd12_run1year_short compare ibased33 missing-data
MISS conda_macos_restart_col_1x1_fsd12_short compare ibased33 missing-data

    - [ ] bit for bit
    - [x] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

**Snow parameter**:  Oleson et al eq 3.68:
> dr_e/dt = 10^18 * C1 * f_liq^3 / (4 * pi * r_e^2)
where  
C1 = 4.22e-13 and f_liq = mass fraction of liquid in snow.  

This PR removes a factor of 100 from f_liq in the current code and changes the constant to S_wet = 10^18 * C1 = 4.22e5. 

**Snow drainage**: this change allows wet metamorphism to function when liquid water associated with the snow scheme is not used directly in melt ponds, i.e. `use_smliq_pnd = .false.`  It also fixes a bug in the computation of drained water mass.

**Brine volume**:  this change is only to the before-and-after calculations used to check brine volume conservation.  There is an extra factor of 10 which ought to be removed, if possible, but I'm leaving it in for now.

All bugs were found during the E3SM/Icepack merge process.

### DIFF
--- a/columnphysics/icepack_snow.F90
+++ b/columnphysics/icepack_snow.F90
@@ -31,7 +31,8 @@
 
       real (kind=dbl_kind), parameter, public :: &
          S_r  = 0.033_dbl_kind, & ! irreducible saturation (Anderson 1976)
-         S_wet= 4.22e-5_dbl_kind  ! (um^3/s) wet metamorphism parameters
+         S_wet= 4.22e5_dbl_kind  ! wet metamorphism parameter (um^3/s)
+                                  ! = 1.e18 * 4.22e-13 (Oleson 2010)
 
       real (kind=dbl_kind) :: &
          min_rhos, &   ! snowtable axis data, assumes linear data
@@ -1079,7 +1080,7 @@
       dr_wet = c0
       fliq = c1
       if (smice + smliq > c0 .and. rsnw > c0) then
-         fliq = min(smliq/(smice + smliq),p1)*c100
+         fliq = min(smliq/(smice + smliq),p1)
          dr_wet = S_wet * fliq**3*dt/(c4*pi*rsnw**2)
       endif
 
--- a/columnphysics/icepack_snow.F90
+++ b/columnphysics/icepack_snow.F90
@@ -16,7 +16,7 @@
       use icepack_parameters, only: isnw_T, isnw_Tgrd, isnw_rhos
       use icepack_parameters, only: snowage_rhos, snowage_Tgrd, snowage_T
       use icepack_parameters, only: snowage_tau, snowage_kappa, snowage_drdt0
-      use icepack_parameters, only: snw_aging_table
+      use icepack_parameters, only: snw_aging_table, use_smliq_pnd
 
       use icepack_therm_shared, only: icepack_ice_temperature
       use icepack_therm_shared, only: adjust_enthalpy
@@ -1161,7 +1161,8 @@
 
       real (kind=dbl_kind) :: &
          hslyr,  & ! snow layer thickness (m)
-         hsn       ! snow thickness (m)
+         hsn,    & ! snow thickness (m)
+         sliq      ! snow liquid content (kg/m^2)
 
       real (kind=dbl_kind), dimension(nslyr) :: &
          dlin    , & ! liquid mass into the layer from above (kg/m^2)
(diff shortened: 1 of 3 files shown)


apcraig (approved):

njeffery (approved):
Approved by inspection.  Parameter for wet metamorphism is now consistent with Brun 1989.

apcraig:
Ran the QC, it fails.  The plots suggest the ice is a bit thinner in the new simulation.  Images below are baseline, new modifications, difference.  I tested with CICE cb58257857d429c4c2d and Icepack main plus changes up to commit #236023f above.
ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcbase

ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb
ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcbase_minus_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb

eclare108213:
Thank you, @apcraig.  I would expect the snow grains to be larger and the resulting ice thinner with the parameter change.  The magnitude of the difference (generally less than 0.5 m) does not concern me, especially considering how thick the ice is in the southern hemisphere.  That's a bit surprising.  Out of curiosity, do you have a QC run with just the default parameters, i.e. with snwgrain and snwitdrdg not turned on?

apcraig:
The ice thickness of the baseline run (snw options off) is much thinner than either of the runs with snw options on.  The two snw runs are closer to each other than to the baseline.  I attach figures for the baseline (no snw options), baseline snw configuration, and new snw configuration (i.e. including mods in this PR) as well as one diff, the baseline - new snw configuration.

ice_thickness_cheyenne_intel_qcchk_gx1_144x1_medium_qc_qcchk
ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcbase
ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb
ice_thickness_cheyenne_intel_qcchk_gx1_144x1_medium_qc_qcchk_minus_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb

eclare108213:
Thank you.  Very interesting, and these plots are generating lots more questions.  I will move this conversation into an issue.  In the meantime, I think we should merge this PR to get the bug fixes into main, and then migrate them into the E3SM fork of Icepack.  Anyone opposed?

apcraig:
One important comment I might add.  The snw cases are starting with the "icdefault" option because restart files are not available.  The standard case starts with a proper initial condition.   Also, I'm not sure whether these plots are means or something else, maybe someone else knows.

njeffery:
The magnitudes of the changes look reasonable to me.

njeffery (approved):
Changes to drain snow  also look good.  These should be BFB.

apcraig:
One more data point, ran a QC for the standard run + icdefault.  This is with snw options off.  3 images follow, standard run + icdefault, current snw options run + icdefault (required), difference.  The runs still fail QC but the differences in the thickness are reduced relative to comparisons with the standard run from a restart file.

ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc qcbaseic

ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb


ice_thickness_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc qcbaseic_minus_cheyenne_intel_smoke_gx1_144x1_icdefault_medium_qc_snwgrain_snwitdrdg qcsbb

apcraig:
Will merge this now and create a PR to migrate this to the E3SM-Project branch.

Issue #255, opened 2019-03-13
Icepack Interface Refactor


apcraig:
A number of concerns associated with the icepack interfaces have been raised and an interface refactor is being considered.  A google doc has been created to provide a single location to aggregate input and to facilitate creation of a design document.  I hope the google doc will lead to a design and implementation plan should we decide to move forward.

https://docs.google.com/document/d/1jgNia1OddbfXW99qFw7r_0d8X_ikrB2oXXF--juZC3c

This should be publicly viewable and I will give most of the Consortium Team write access.  Feel free to edit the google doc directly, email me with comments/thoughts separately, or even add comments under this github issue.    I will try to keep the google doc fresh.

eclare108213:
Thanks @apcraig 
For the two proposed approaches (keywords and data types) for mitigating issues with long hardwired subroutine argument lists, I would like to know what experience modeling centers/models have with them.  What is the best way to evaluate these options, short of implementing both and comparing?  Are there other potential solutions?
@philipwjones do you have suggestions this?

apcraig:
Thanks @eclare108213.  Let me also just clarify that the current google doc is something I put together in 30 minutes.  We need to flush out other issues, other requirements, and other design ideas.  The current document is far from comprehensive, but hopefully serves as a starting point.  I hope we will add significant content to the document over the next days/weeks and reorganize the document as needed to meet our needs.

philipwjones:
Sorry to be slow in responding. A few comments on the plan so far.  Despite the length of argument lists, explicit arguments with clear in/out intent is still going to be preferred for the ability to determine dependencies and for new programming models that infer data flow or create a DAG from those dependencies. Long argument lists can be alleviated with the use of user-defined types or structs.  But I would not recommend a large pbuf-like type since we'd still like some granularity in the dependency analysis to determine where some process/task parallelism can be exploited. Also, we'd like very shallow types for both compiler efficiency (avoiding need for deep copies) and because it's easier to extract data without chasing pointers down a deep structure. 

Regarding keywords and optional args - this is a Fortran feature (and a nice one at that!) that can be a risky approach as we continue down the road toward a mixed-language environments and ongoing pressure to abandon Fortran. Something to consider since Icepack needs to be adopted by newer C-based ice codes in the pipeline and long-term support for Fortran continues to decline.

eclare108213:
See description of interface changes needed by GFDL here:
[...]

PR #423, opened 2023-01-11
Refactor optional argument implementation for isotopes, snwgrain, therm1 and therm2

apcraig:
## PR checklist
    Refactor isotope implementation with respect to optional arguments
- [X] Developer(s): 
    apcraig
    CICE is bit-for-bit on cheyenne, full test suite.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#0bf0fdcabd00df95d1028a66f71784b0b68f1178.  Icepack regression testing bit-for-bit on cheyenne https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#0c573347b261eba1ab6a373273cc232a7feb1ee1
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- Use argcheck design
- Eliminate local copies of optional data
- Pass optional arguments down the calling tree
- Some minor format cleanup

### DIFF
--- a/columnphysics/icepack_atmo.F90
+++ b/columnphysics/icepack_atmo.F90
@@ -61,7 +61,6 @@
                                       Cdn_atm,            &
                                       Cdn_atm_ratio_n,    &
                                       Qa_iso,   Qref_iso, &
-                                      iso_flag,           &
                                       uvel,     vvel,     &
                                       Uref,     zlvs      )
 
@@ -103,13 +102,10 @@
          shcoef   , & ! transfer coefficient for sensible heat
          lhcoef       ! transfer coefficient for latent heat
 
-      logical (kind=log_kind), intent(in), optional :: &
-         iso_flag     ! flag to trigger iso calculations
-
-      real (kind=dbl_kind), intent(in), optional, dimension(:) :: &
+      real (kind=dbl_kind), intent(in), dimension(:), optional :: &
          Qa_iso       ! specific isotopic humidity (kg/kg)
 
-      real (kind=dbl_kind), intent(inout), optional, dimension(:) :: &
+      real (kind=dbl_kind), intent(inout), dimension(:), optional :: &
          Qref_iso     ! reference specific isotopic humidity (kg/kg)
 
       real (kind=dbl_kind), intent(in) :: &
@@ -167,15 +163,7 @@
       real (kind=dbl_kind), parameter :: &
          zTrf  = c2     ! reference height for air temp (m)
 
-      logical (kind=log_kind) :: &
-         l_iso_flag     ! local flag to trigger iso calculations
-
       character(len=*),parameter :: subname='(atmo_boundary_layer)'
-
-      l_iso_flag = .false.
-      if (present(iso_flag)) then
-        l_iso_flag = iso_flag
-      endif
 
       al2 = log(zref/zTrf)
 
@@ -389,21 +377,21 @@
          Uref = vmag * rd / rdn
       endif
 
-      if (l_iso_flag) then
-       if (present(Qref_iso) .and. present(Qa_iso)) then
+      if (present(Qref_iso)) then
          Qref_iso(:) = c0
          if (tr_iso) then
-            do n = 1, n_iso
-               ratio = c0
-               if (Qa_iso(2) > puny) ratio = Qa_iso(n)/Qa_iso(2)
-               Qref_iso(n) = Qa_iso(n) - ratio*delq*fac
-            enddo
+            if (present(Qa_iso)) then
+               do n = 1, n_iso
+                  ratio = c0
+                  if (Qa_iso(2) > puny) ratio = Qa_iso(n)/Qa_iso(2)
+                  Qref_iso(n) = Qa_iso(n) - ratio*delq*fac
+               enddo
+            else
+               call icepack_warnings_add(subname//' Qref_iso and Qa_iso both must be passed with tr_iso')
+               call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
+               return
+            endif
          endif
-       else
-         call icepack_warnings_add(subname//' l_iso_flag true but optional arrays missing')
-         call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
-         return
-       endif
       endif
 
       end subroutine atmo_boundary_layer
@@ -895,18 +883,18 @@
          shcoef   , & ! transfer coefficient for sensible heat
          lhcoef       ! transfer coefficient for latent heat
 
-      real (kind=dbl_kind), intent(in), optional, dimension(:) :: &
+      real (kind=dbl_kind), intent(in), dimension(:), optional :: &
          Qa_iso       ! specific isotopic humidity (kg/kg)
 
-      real (kind=dbl_kind), intent(inout), optional, dimension(:) :: &
+      real (kind=dbl_kind), intent(inout), dimension(:), optional :: &
          Qref_iso     ! reference specific isotopic humidity (kg/kg)
 
-      real (kind=dbl_kind), optional, intent(in) :: &
+      real (kind=dbl_kind), intent(in), optional :: &
          uvel     , & ! x-direction ice speed (m/s)
          vvel     , & ! y-direction ice speed (m/s)
          zlvs         ! atm level height for scalars (if different than zlvl) (m)
 
-      real (kind=dbl_kind), optional, intent(out) :: &
+      real (kind=dbl_kind), intent(out), optional :: &
          Uref         ! reference height wind speed (m/s)
 
 !autodocument_end
@@ -915,12 +903,6 @@
 
       real (kind=dbl_kind) :: &
          l_uvel, l_vvel, l_Uref
-
-      real (kind=dbl_kind), dimension(:), allocatable :: &
-         l_Qa_iso, l_Qref_iso   ! local copies of Qa_iso, Qref_iso
-
-      logical (kind=log_kind) :: &
-         iso_flag  ! flag to turn on iso calcs in other subroutines
 
       character(len=*),parameter :: subname='(icepack_atm_boundary)'
 
@@ -932,19 +914,6 @@
       endif
       if (present(vvel)) then
          l_vvel = vvel
-      endif
-      if (present(Qa_iso) .and. present(Qref_iso)) then
-         iso_flag = .true.
-         allocate(l_Qa_iso(size(Qa_iso,dim=1)))
-         allocate(l_Qref_iso(size(Qref_iso,dim=1)))
-         l_Qa_iso = Qa_iso
-         l_Qref_iso = Qref_iso
-      else
-         iso_flag = .false.
-         allocate(l_Qa_iso(1))
-         allocate(l_Qref_iso(1))
-         l_Qa_iso = c0
-         l_Qref_iso = c0
       endif
 
       Cdn_atm_ratio_n = c1
@@ -972,23 +941,16 @@
                                    lhcoef,   shcoef,        &
                                    Cdn_atm,                 &
                                    Cdn_atm_ratio_n,         &
-                                   iso_flag = iso_flag,     &
-                                   Qa_iso=l_Qa_iso,         &
-                                   Qref_iso=l_Qref_iso,     &
-                                   uvel=l_uvel, vvel=l_vvel,  &
-                                   Uref=l_Uref, zlvs=zlvs)
+                                   Qa_iso=Qa_iso,           &
+                                   Qref_iso=Qref_iso,       &
+                                   uvel=l_uvel, vvel=l_vvel,&
+                                   Uref=l_Uref, zlvs=zlvs   )
          if (icepack_warnings_aborted(subname)) return
       endif ! atmbndy
 
       if (present(Uref)) then
          Uref = l_Uref
       endif
-
-      if (present(Qref_iso)) then
-         Qref_iso = l_Qref_iso
-      endif
-
-      deallocate(l_Qa_iso,l_Qref_iso)
 
       end subroutine icepack_atm_boundary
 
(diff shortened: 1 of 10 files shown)


apcraig:
This refactors the isotope implementation using the newer design, passing optional arguments down the calling tree and using the argcheck design.  If this seems reasonable, I will refactor other Icepack features in a similar way where reasonable.  Do we want a PR for each feature refactor or just one (or a few) PRs to cover multiple feature refactors?  Would like to get some feedback before moving forward.

This should be bit-for-bit and has no impact on the public interfaces.  Will run comprehensive tests to confirm TBD.

dabail10 (approved):
Sorry for not responding to the email. This looks fine to me.

columnphysics/icepack_atmo.F90
```diff
@@ -972,24 +941,17 @@ subroutine icepack_atm_boundary(sfctype,                   &
                                    lhcoef,   shcoef,        &
                                    Cdn_atm,                 &
                                    Cdn_atm_ratio_n,         &
-                                   iso_flag = iso_flag,     &
-                                   Qa_iso=l_Qa_iso,         &
-                                   Qref_iso=l_Qref_iso,     &
-                                   uvel=l_uvel, vvel=l_vvel,  &
-                                   Uref=l_Uref, zlvs=zlvs)
+                                   Qa_iso=Qa_iso,           &
+                                   Qref_iso=Qref_iso,       &
```
eclare108213, line 945: What happens if present(Qref_iso)=F?  Is this robust across all compilers/platforms?  You probably tested that previously...

columnphysics/icepack_mechred.F90
```diff
@@ -1825,33 +1826,41 @@ subroutine icepack_step_ridge (dt,           ndtd,         &
...
 
+      !-----------------------------------------------------------------
+      ! Check optional arguments
+      !-----------------------------------------------------------------
+
+      if (argcheck == 'always' .or. (argcheck == 'first' .and. first_call)) then
+         if (tr_iso) then
+            if (present(fiso_ocn)) then
+                ! OK
+            else
```
eclare108213, line 1848: switch the logic to `if (.not.(present(tr_iso))) then` ?

eclare108213:
> Do we want a PR for each feature refactor or just one (or a few) PRs to cover multiple feature refactors?

One or a few to cover multiple features would be fine, but if something is particularly complicated, then just do that by itself.  E.g. I imagine that the thermodynamics will be complicated, since there are many options that feed into it.

columnphysics/icepack_atmo.F90
apcraig: This is robust.  The optional attribute and "present" state is passed down the calling tree.  As long as optional is set in the calling routine for this field if want to check it's "present" state (determined by the original caller) and/or the optional argument is not used unless it exists, all should be fine.  You can definitely get into trouble if you pass the optional argument down the calling tree and use it if it wasn't passed.  Hopefully, we'll avoid that with the logic we have setup.

columnphysics/icepack_mechred.F90
apcraig: Maybe.  But if you need 4 optional arguments, I think this syntax is clearer.  For example,

```
if (present(x1) .and. present(x2). and. present(x3).and.present(x4)) then
  !OK
else
  abort
endif
```

rather than

```
if (.not.present(x1) .or..not.present(x2).or..not.present(x3).or..not.present(x4)) then
  abort
endif
```

I prefer the former.  I know it's awkward too but I think it's cleaner.  I guess a question is whether

```
if (.not.(present(x1).and.present(x2).and.present(x3).and.present(x4))) then
  abort
endif
```

this would also work and maybe be best?  Happy for some input.

apcraig:
I think it may be worth reviewing and merging this now.  Most of the therm1 and therm2 optional arguments have been refactored.  Still waiting on some additional insight about meltsliq, but that can be done in a separate PR.

columnphysics/icepack_mechred.F90
apcraig: I ended up updating it using the final option above.

apcraig:
Icepack and CICE test suites are bit-for-bit.

columnphysics/icepack_atmo.F90
```diff
@@ -904,8 +896,28 @@ subroutine icepack_atm_boundary(sfctype,                   &
...
 
+      logical (kind=log_kind), save :: &
+         first_call = .true.   ! first call flag
+
       character(len=*),parameter :: subname='(icepack_atm_boundary)'
 
+      !------------------------------------------------------------
+      ! Check optional arguments
+      ! This is used for ocean and ice sfctype, need to be careful
+      !------------------------------------------------------------
```
phil-blain, line 907: I'm not sure I understand this comment, why does it say 

> This is used for ocean and ice sfctype

but then we just check `sfctype == 'ice'` ? What is "This" referring to?

phil-blain:
> This refactors the isotope implementation using the newer design, passing optional arguments down the calling tree and using the argcheck design. If this seems reasonable, I will refactor other Icepack features in a similar way where reasonable. 

I think this is a nice clean up, it improves readability,  I think. Also, gotta love a PR that removes more lines of code than it adds ;)

> Do we want a PR for each feature refactor or just one (or a few) PRs to cover multiple feature refactors? Would like to get some feedback before moving forward.

If I were doing it, I think I would probably do it that way:
- 1 PR per module (so 1 per file)
- In each PR, one commit per subroutine: removing `l_*` local variable, using `argcheck` to validate arguments, and adjusting calls to other subroutines if need be.

This would be very much more pleasant to review, at least for me :) 

Also I would avoid changing the order of the variables attributes (like moving `optional` to be the last attributes, with no other changes). This adds a lot of noise to the diffs. If we want to make such changes, that's OK, but I'd prefer if it were done in separate commits.

Thanks a lot for starting the work on this! :)

columnphysics/icepack_atmo.F90
```diff
@@ -916,14 +896,28 @@ subroutine icepack_atm_boundary(sfctype,                   &
...
 
       character(len=*),parameter :: subname='(icepack_atm_boundary)'
 
+      !------------------------------------------------------------
+      ! Check optional arguments
+      ! This is used for ocean and ice sfctype, need to be careful
+      !------------------------------------------------------------
+
+      if (sfctype == 'ice') then
+      if (argcheck == 'always' .or. (argcheck == 'first' .and. first_call_ice)) then
```
phil-blain, line 910: The logic in this line is duplicated a few times. Would it make sense to isolate it to a function somewhere ? Then we could `use` that function instead of the `argcheck` variable.

The function could receive that boolean `fist_call` (or here `first_call_ice`) as argument and return a boolean...

columnphysics/icepack_snow.F90
```diff
@@ -1194,6 +1194,7 @@ subroutine drain_snow (nslyr, vsnon, aicen, &
       else
          sliq = meltsliq  ! computed in thickness_changes
       endif
+!tcx??
       meltsliq = meltsliq
```
eclare108213, line 1198: This line can be commented out.  This section of the code was rewritten a bit and corrected, and this line was the logical culmination of that process.

columnphysics/icepack_therm_itd.F90
```diff
@@ -1852,15 +1853,15 @@ subroutine add_new_ice (ncat,      nilyr,      &
                enddo
             endif
 
-           frazil_conc = c0
```
eclare108213, line 1855: This line is still present in a similar section of code, above.  The code is hard-wired for n_iso=3, apparently, so maybe this line should stay in case for some reason n_iso > 3?

columnphysics/icepack_therm_vertical.F90
```diff
@@ -2381,178 +2382,92 @@ subroutine icepack_step_therm1(dt, ncat, nilyr, nslyr,    &
...
+               call icepack_warnings_add(subname//' error in snw arguments, snwgrain=T')
+               call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
+               return
+            endif
+         endif
+         if ((present(fswthru_vdr) .and. .not.present(fswthrun_vdr)) .or. &
+             (present(fswthru_vdf) .and. .not.present(fswthrun_vdf)) .or. &
+             (present(fswthru_idr) .and. .not.present(fswthrun_idr)) .or. &
+             (present(fswthru_idf) .and. .not.present(fswthrun_idf))) then
+            call icepack_warnings_add(subname//' error in fswthru [iv]d[rf] arguments')
```
eclare108213, line 2445: @apcraig @dabail10  Is this the logic we want here?  Often it's fine to have category-aggregated variables and not keep the category-specific ones.

eclare108213 (commented):
Just a few comments and questions -

columnphysics/icepack_atmo.F90
apcraig: I need to clean up this comment.

apcraig:
> Just a few comments and questions -
> 
> I do worry a bit about Phil's comment [here](https://github.com/CICE-Consortium/Icepack/issues/255#issuecomment-508872338), regarding non-Fortran codes being able to use Icepack with Fortran keywords and optional arguments. What should we do about that?

One option is for the non-Fortran interfaces to NOT adopt the optional arguments in their interfaces/uses.  So non-Fortran might pass all arguments even if they are not needed.  I don't actually know whether that will work for all languages, but we should explore as an appropriate time.  I think for now, we should continue to add optional arguments moving forward.  It provide backwards compatibility in the current community.  If at some point in the future, we decide to not have optional arguments in icepack, we just remove the optional attribute on the Icepack interfaces and then update any interfaces in CICE/Icepack that do not conform.  Everyone in the community would have to do the same, also anytime Icepacks is updated with any new arguments.  That's the downside overall.  I don't think we should abandon optional arguments until there is a clear reason to do so.

columnphysics/icepack_therm_vertical.F90
apcraig: It's really the underlying implementation.  The implementation in Icepack with this update will always computed the aggregated variables if they are passed, and that requires the category specific arguments too.  We could change the implementation so both have to be passed for the category aggregate to be computed.  The optional argument checking is determined to a large degree on the underlying implementation.

columnphysics/icepack_atmo.F90
```diff
@@ -389,21 +377,13 @@ subroutine atmo_boundary_layer (sfctype,            &
...
-         endif
-       else
-         call icepack_warnings_add(subname//' l_iso_flag true but optional arrays missing')
-         call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
-         return
-       endif
+         do n = 1, n_iso
+            ratio = c0
+            if (Qa_iso(2) > puny) ratio = Qa_iso(n)/Qa_iso(2)
+            Qref_iso(n) = Qa_iso(n) - ratio*delq*fac
```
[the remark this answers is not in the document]
apcraig: That's right, we have already checked the arguments in the highest level Icepack interface.  So they are safe to use here without checking.  That's our current design.  As I mentioned the other day, there are pros and cons to doing it this way.  A pro is that we check all the optional arguments early leveraging parameter flags and then we should be OK after that without having to check over and over.  A negative is that you have to keep track of what's going on in different layers of the code.

columnphysics/icepack_snow.F90
apcraig: > > Do we want a PR for each feature refactor or just one (or a few) PRs to cover multiple feature refactors? Would like to get some feedback before moving forward.
> 
> If I were doing it, I think I would probably do it that way:
> 
>     * 1 PR per module (so 1 per file)
> 
>     * In each PR, one commit per subroutine: removing `l_*` local variable, using `argcheck` to validate arguments, and adjusting calls to other subroutines if need be.

I don't think 1 PR per module will work well.  The optional argument checking will depend on the underlying implementation and that needs to be coordinated.  Removing the l_ variables might result in further optimizations down the calling tree.  So maybe 1 PR per feature.  I also understand about general cleanup (formatting) with updating features.  I agree, it's just hard not to fix things when I see an opportunity, and I'm worried I may not circle back if I don't.  I'll try to be more careful.

eclare108213:
> > Just a few comments and questions -
> > I do worry a bit about Phil's comment [here](https://github.com/CICE-Consortium/Icepack/issues/255#issuecomment-508872338), regarding non-Fortran codes being able to use Icepack with Fortran keywords and optional arguments. What should we do about that?
> 
> One option is for the non-Fortran interfaces to NOT adopt the optional arguments in their interfaces/uses. So non-Fortran might pass all arguments even if they are not needed. I don't actually know whether that will work for all languages, but we should explore as an appropriate time. I think for now, we should continue to add optional arguments moving forward. It provide backwards compatibility in the current community. If at some point in the future, we decide to not have optional arguments in icepack, we just remove the optional attribute on the Icepack interfaces and then update any interfaces in CICE/Icepack that do not conform. Everyone in the community would have to do the same, also anytime Icepacks is updated with any new arguments. That's the downside overall. I don't think we should abandon optional arguments until there is a clear reason to do so.

The voice of reason - sounds good to me.

columnphysics/icepack_atmo.F90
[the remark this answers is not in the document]
eclare108213: Okay, thank you.

eclare108213:
Looks like there are just a few minor changes that could be made here.  @apcraig let us know when you're ready to merge.

apcraig:
cheyenne was down this week and i'm away from home, should be able to make progress next few days.  will update conservation when updates are ready.

apcraig:
I've updated the PR and running a full test suite now.  Added icepack_chkoptargflag logical function and cleaned up a few other things associated with comments above.  Feel free to review, will report updated results in the next day or so.

apcraig:
CICE testing is complete with latest changes and everything looks good.

phil-blain (approved):
Thanks, latest version looks good!

eclare108213 (approved):

PR #426, opened 2023-01-26
driver: icedrv_history: fix start time for NetCDF history output

davidclemenssewall:

## PR checklist
- [X ] Short (1 sentence) summary of your PR: 
    The start time for netCDF output was hardcoded to be at 00:00:00 and there was a bug in icedrv_calendar such that idate0 was fixed at Jan. 1. Both bugs are fixed.
- [X ] Developer(s): 
    David Clemens-Sewall
- [X ] Suggest PR reviewers from list in the column to the right.
- Tony Craig, Dave Bailey 
- [X ] Please copy the PR test results link or provide a summary of testing completed below.
    Icepack base_suite 128/128 passed. CICE quick_suite 16/16 passed.
    - [X ] bit for bit (except for origin of time dimension in netcdf output)
    - [ ] more substantial 
    - [ ] Yes
    - [X ] No
    - [ ] Yes
    - [X ] No
    - [ ] Yes
    - [X ] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [X ] No 

### DIFF
--- a/configuration/driver/icedrv_calendar.F90
+++ b/configuration/driver/icedrv_calendar.F90
@@ -167,7 +167,7 @@
       tday = (time-sec)/secday + c1     ! absolute day number
 
       ! Convert the current timestep into a calendar date
-      call sec2time(nyr,month,mday,basis_seconds+sec)
+      call sec2time(nyr,month,mday,basis_seconds+time)
 
       yday = mday + daycal(month)  ! day of the year
       nyr = nyr - year_init + 1    ! year number
--- a/configuration/driver/icedrv_calendar.F90
+++ b/configuration/driver/icedrv_calendar.F90
@@ -61,7 +61,6 @@
          year_init, & ! initial year
          nyr      , & ! year number
          idate    , & ! date (yyyymmdd)
-         idate0   , & ! initial date (yyyymmdd)
          sec      , & ! elapsed seconds into date
          npt      , & ! total number of time steps (dt)
          ndtd     , & ! number of dynamics subcycles: dt_dyn=dt/ndtd
@@ -171,8 +170,6 @@
 
       yday = mday + daycal(month)  ! day of the year
       nyr = nyr - year_init + 1    ! year number
-
-      idate0 = (nyr+year_init-1)*10000 + month*100 + mday ! date (yyyymmdd)
 
       end subroutine init_calendar
 
--- a/configuration/driver/icedrv_history.F90
+++ b/configuration/driver/icedrv_history.F90
@@ -68,7 +68,10 @@
          count1(1), count2(2), count3(3), count4(4), & ! cdf start/count arrays
          varid, &                        ! cdf varid
          status, &                       ! cdf status flag
-         iflag                           ! history file attributes
+         iflag, &                        ! history file attributes
+         h0, &                           ! start hour
+         m0, &                           ! start minute
+         s0                              ! start second
 
       character (len=8) :: &
          cdate                           ! date string
@@ -158,6 +161,9 @@
          endif
 
          ! time dimension
+         h0 = sec / 3600 ! Get the current hour
+         m0 = mod(sec, 3600) / 60 ! Get the current minute
+         s0 = mod(sec, 60) ! Get the current seconds
          status = nf90_def_dim(ncid,'time',NF90_UNLIMITED,timid)
          if (status /= nf90_noerr) call icedrv_system_abort(string=subname//' ERROR: def_dim time')
          status = nf90_def_var(ncid,'time',NF90_DOUBLE,timid,varid)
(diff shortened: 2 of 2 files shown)


apcraig:
@davidclemenssewall, I'm curious about this fix.  What was the problem exactly?  Was the netcdf time axis off when starting at some time other than 00:00:00?  If that's the case, I think a clearer fix would be to leave the netcdf time axis as before, relative to 00:00:00, but just update the calculation of the time values on the netcdf time axis.  Does that make sense?

davidclemenssewall:
@apcraig There were two issues here and the symptom of both of them was that NetCDF history output was always labelled as starting at YYYY-01-01 01:00:00 regardless of what value was set for istep0. For example, if we set istep0 = 4320, then the first time step should be YYYY-06-01 01:00:00. However, the netcdf history file created would still indicate that it started Jan. 01. If we look at the actual data values in that file it appears that it is correctly starting on Jun. 30 (e.g., fsw is not zero).

I've implemented the time value calculation as you've suggested (see most recent commit). Note that with this change the idate0 variable is actually no longer used anywhere in the code and it was kind of misleading anyways so I've removed it. Also, for the netCDF history files the year was restricted to be a four digit number, which didn't seem to have any purpose (netCDF and cftime don't have this restriction). So I removed that restriction as well.

apcraig:
@davidclemenssewall, thanks for the latest updates.  These kind of changes always scare me.  The calendar is very delicate in general.  In fact, the CICE time manager was refactored a while ago, but those changes were not pushed back into the icepack driver.  How much testing have you done?  I will also run a full test suite just to make sure everything still runs fine.

apcraig:
Full test suite on cheyenne passes bit-for-bit, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#e0560169c439a9aaf3a3f2024e42f7d6a009fd1f.

apcraig (approved):

davidclemenssewall:
@apcraig I ran the Icepack base_suite and the CICE quick_suite (both of which all passed). I also did a number of sanity checks of what time range output should cover with different values of istep0 and npt. The changes in icedrv_history only impact the history output, so there shouldn't be any risk of them impacting the rest of the calendar. 

In icedrv_calendar, the only change is to cause this line (`call sec2time(nyr,month,mday,basis_seconds+time)`) to actually have the effect that the comment above it says that it should. That is, before changing 'sec' to 'time' in the final argument that line always set month and mday to both be 1, regardless of the value of istep0. I don't think fixing that could damage the calendar, but I'm only just getting to know this code base so I'm not sure what to check.

apcraig:
I know the test suite runs fine.  Just to confirm that you tested some cases where the start date/istep/npt were offset and you've confirmed this fix works fine in those cases too?

davidclemenssewall:
I tried a couple different combinations of istep0 and npt and confirmed that the history files started at the right times and had the correct number of rows. I also spot checked some of the forcing variables (e.g., flw) to make sure that they looked the same at the same time of year independent of when istep0 was.

configuration/driver/icedrv_calendar.F90
```diff
@@ -167,13 +166,11 @@ subroutine init_calendar
       tday = (time-sec)/secday + c1     ! absolute day number
 
       ! Convert the current timestep into a calendar date
-      call sec2time(nyr,month,mday,basis_seconds+sec)
+      call sec2time(nyr,month,mday,basis_seconds+time)
```
dabail10, line 169: This seems strange to me. The variable time should be the time since the beginning of the run. However, I see that this is upon initialization and that later in the calendar subroutine, the call to sec2time needs the time since the first step. So, I think this is correct.

configuration/driver/icedrv_history.F90
```diff
@@ -68,7 +68,11 @@ subroutine history_write()
          count1(1), count2(2), count3(3), count4(4), & ! cdf start/count arrays
          varid, &                        ! cdf varid
          status, &                       ! cdf status flag
-         iflag                           ! history file attributes
+         iflag, &                        ! history file attributes
+         sec0, &                         ! number of seconds into the day at istep0
+         h0, &                           ! start hour
+         m0, &                           ! start minute
+         s0                              ! start second
```
dabail10, line 75: Are these actually used?

configuration/driver/icedrv_history.F90
```diff
@@ -164,9 +168,8 @@ subroutine history_write()
          if (status /= nf90_noerr) call icedrv_system_abort(string=subname//' ERROR: def_var time')
          status = nf90_put_att(ncid,varid,'long_name','model time')
          if (status /= nf90_noerr) call icedrv_system_abort(string=subname//' ERROR: put_att time long_name')
-         write(cdate,'(i8.8)') idate0
-         write(tmpstr,'(a,a,a,a,a,a,a,a)') 'days since ', &
-            cdate(1:4),'-',cdate(5:6),'-',cdate(7:8),' 00:00:00'
+         write(tmpstr,'(a,i0,a)') 'days since ', &
+            year_init,'-01-01 00:00:00'
```
dabail10, line 172: We have to be careful here. I guess it is ok for the Icepack driver. idate0 was used here to correspond to the CICE driver for history.

apcraig:
The CICE time manager was rewritten a while ago, so it's different now.  I think any fixes like this in the Icepack driver are probably OK now as then are no longer connected to coupling.

configuration/driver/icedrv_history.F90
davidclemenssewall: Ah good catch. Those are not actually used. They were used in the original way that I had fixed this. But since we've decided that the start point will always be January 1 00:00:00 and we just use the time variable they are not needed. I'll remove them in the NetCDF restart branch.

PR #429, opened 2023-02-14
Update implementation of some optional arguments in Icepack

apcraig:
## PR checklist
    Update implementation of some optional arguments in Icepack
- [X] Developer(s): 
    apcraig
    Full test suite on cheyenne with 3 compilers with both Icepack, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#67c874194a4acdaf09fc756cd679f579fa53060d, and CICE, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#b946a95ea46096d739b4b0d725d0eab81a53fbee are bit-for-bit.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Remove local copies where possible

Check optional arguments 

Remove public interface declarations where not needed 

Clean up some intent statements

### DIFF
--- a/columnphysics/icepack_fsd.F90
+++ b/columnphysics/icepack_fsd.F90
@@ -120,17 +120,9 @@
       real (kind=dbl_kind), dimension(:), allocatable :: &
          lims
 
-      logical (kind=log_kind) :: &
-         l_write_diags  ! local write diags
-
       character(len=8) :: c_fsd1,c_fsd2
       character(len=2) :: c_nf
       character(len=*), parameter :: subname='(icepack_init_fsd_bounds)'
-
-      l_write_diags = .true.
-      if (present(write_diags)) then
-         l_write_diags = write_diags
-      endif
 
       if (nfsd.eq.24) then
 
@@ -230,7 +222,8 @@
          c_fsd_range(n)=c_fsd1//'m < fsd Cat '//c_nf//' < '//c_fsd2//'m'
       enddo
 
-      if (l_write_diags) then
+      if (present(write_diags)) then
+      if (write_diags) then
          write(warnstr,*) ' '
          call icepack_warnings_add(warnstr)
          write(warnstr,*) subname
@@ -243,6 +236,7 @@
          enddo
          write(warnstr,*) ' '
          call icepack_warnings_add(warnstr)
+      endif
       endif
 
       end subroutine icepack_init_fsd_bounds
--- a/columnphysics/icepack_orbital.F90
+++ b/columnphysics/icepack_orbital.F90
@@ -177,11 +177,24 @@
 
       real (kind=dbl_kind) :: ydayp1 ! day of year plus one time step
 
+      logical (kind=log_kind), save :: &
+         first_call = .true.   ! first call flag
+
       character(len=*),parameter :: subname='(compute_coszen)'
 
 ! Solar declination for next time step
 
 #ifdef CESMCOUPLED
+      if (icepack_chkoptargflag(first_call)) then
+         if (.not.(present(days_per_year) .and. &
+                   present(nextsw_cday)   .and. &
+                   present(calendar_type))) then
+            call icepack_warnings_add(subname//' error in CESMCOUPLED args')
+            call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
+            return
+         endif
+      endif
+
       if (calendar_type == "GREGORIAN") then
          ydayp1 = min(nextsw_cday, real(days_per_year,kind=dbl_kind))
       else
@@ -205,6 +218,8 @@
 #ifdef CESMCOUPLED
       endif
 #endif
+
+      first_call = .false.
 
       end subroutine compute_coszen
 
--- a/columnphysics/icepack_shortwave.F90
+++ b/columnphysics/icepack_shortwave.F90
@@ -66,10 +66,7 @@
       implicit none
 
       private
-      public :: run_dEdd, &
-                shortwave_ccsm3, &
-                compute_shortwave_trcr, &
-                icepack_prep_radiation, &
+      public :: icepack_prep_radiation, &
                 icepack_step_radiation
 
       real (kind=dbl_kind), parameter :: &
@@ -178,22 +175,17 @@
          alvdfns, & ! visible, diffuse, snow  (fraction)
          alidfns    ! near-ir, diffuse, snow  (fraction)
 
-      real (kind=dbl_kind), dimension(:), allocatable :: &
-         l_fswthru_vdr  , & ! vis dir SW through ice to ocean (W m-2)
-         l_fswthru_vdf  , & ! vis dif SW through ice to ocean (W m-2)
-         l_fswthru_idr  , & ! nir dir SW through ice to ocean (W m-2)
-         l_fswthru_idf      ! nir dif SW through ice to ocean (W m-2)
+      real (kind=dbl_kind) :: &
+         l_fswthru_vdr, & ! vis dir SW through ice to ocean (W m-2)
+         l_fswthru_vdf, & ! vis dif SW through ice to ocean (W m-2)
+         l_fswthru_idr, & ! nir dir SW through ice to ocean (W m-2)
+         l_fswthru_idf    ! nir dif SW through ice to ocean (W m-2)
 
       character(len=*),parameter :: subname='(shortwave_ccsm3)'
 
       !-----------------------------------------------------------------
       ! Solar radiation: albedo and absorbed shortwave
       !-----------------------------------------------------------------
-
-      allocate(l_fswthru_vdr(ncat))
-      allocate(l_fswthru_vdf(ncat))
-      allocate(l_fswthru_idr(ncat))
-      allocate(l_fswthru_idf(ncat))
 
       ! For basic shortwave, set coszen to a constant between 0 and 1.
       coszen = p5 ! sun above the horizon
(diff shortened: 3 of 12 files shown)


apcraig:
I think I've reviewed all the current optional arguments in Icepack now and checked implementation.  There are probably other Icepack arguments that could be made optional, but we'll need to come up with some suggestions if we want to do that.

I will merge this to the E3SM Icepack branch after it's merged and update CICE as well.

columnphysics/icepack_orbital.F90
```diff
@@ -177,11 +177,24 @@ subroutine compute_coszen (tlat,          tlon,        &
...
+      if (icepack_chkoptargflag(first_call)) then
+         if (.not.(present(days_per_year) .and. &
+                   present(nextsw_cday)   .and. &
+                   present(calendar_type))) then
+            call icepack_warnings_add(subname//' error in CESMCOUPLED args')
+            call icepack_warnings_setabort(.true.,__FILE__,__LINE__)
+            return
+         endif
+      endif
+
```
phil-blain, line 197: OK, so if I understand correctly these flags are not optional for CESM, so we add a check to that effect here?

If so, that makes sense.

columnphysics/icepack_shortwave.F90
```diff
@@ -295,29 +287,24 @@ subroutine shortwave_ccsm3 (aicen,    vicen,    &
...
                                fswint=fswint(n),     &
                                fswthru=fswthru(n),   &
-                               fswthru_vdr=l_fswthru_vdr(n),&
-                               fswthru_vdf=l_fswthru_vdf(n),&
-                               fswthru_idr=l_fswthru_idr(n),&
-                               fswthru_idf=l_fswthru_idf(n),&
+                               fswthru_vdr=l_fswthru_vdr,&
+                               fswthru_vdf=l_fswthru_vdf,&
+                               fswthru_idr=l_fswthru_idr,&
+                               fswthru_idf=l_fswthru_idf,&
```
phil-blain, line 293: OK, here we are in the call to `absorbed_solar` (not enough context is shown in this diff), and in that subroutine `fswthru_vdr` and friends are not optional, so here we have to keep the local counterparts. 

It makes sense to makes these scalars instead, nice cleanup.

columnphysics/icepack_shortwave.F90
```diff
@@ -295,29 +287,24 @@ subroutine shortwave_ccsm3 (aicen,    vicen,    &
...
                                fswpenl=fswpenl(:,n), &
                                Iswabs=Iswabs(:,n))
 
          if (icepack_warnings_aborted(subname)) return
 
+         if(present(fswthru_vdr)) fswthru_vdr(n) = l_fswthru_vdr
+         if(present(fswthru_vdf)) fswthru_vdf(n) = l_fswthru_vdf
+         if(present(fswthru_idr)) fswthru_idr(n) = l_fswthru_idr
+         if(present(fswthru_idf)) fswthru_idf(n) = l_fswthru_idf
```
phil-blain, line 302: This is a very nice cleanup because these statements are inside the condition on `aicen`. So this means that, if for some category the condition is not matched, whatever values were in `fswthru_idf` and friends stay there untouched. This was not the case before, since the whole arrays were overwritten by their local counterparts, which themselves were uninitialized. This lead to a crash in one of our configurations (when compiling with NaN initialization) which lead me to carry this commit in our internal fork: https://github.com/phil-blain/Icepack/commit/c4776379d95501d90ae79fd86d234472cd02d4f9

Nice to know I'll be able to drop this one eventually :)

columnphysics/icepack_therm_vertical.F90
```diff
@@ -2457,11 +2455,10 @@ subroutine icepack_step_therm1(dt, ncat, nilyr, nslyr,    &
       smice(:) = c0
       smliq(:) = c0
 
-      allocate(l_meltsliqn(ncat))
-      l_meltsliqn = c0
-      if (present(meltsliqn)) l_meltsliqn = meltsliqn
       l_meltsliq  = c0
+      l_meltsliqn = c0
```
phil-blain, line 2459: OK, this was already initialized below at:
https://github.com/CICE-Consortium/Icepack/blob/67c874194a4acdaf09fc756cd679f579fa53060d/columnphysics/icepack_therm_vertical.F90#L2548-L2555

but I think it makes sense to initialize it here, closer to the beginning of the subroutine. Maybe we'd want to remove that other initialization then ?

columnphysics/icepack_shortwave.F90
```diff
@@ -66,10 +66,7 @@ module icepack_shortwave
       implicit none
 
       private
-      public :: run_dEdd, &
-                shortwave_ccsm3, &
-                compute_shortwave_trcr, &
-                icepack_prep_radiation, &
```
phil-blain, line 72: were the interfaces of these subroutines documented via the autodocumentation script (I did not check) ? if so, ww would want to re-run it to remove them from the public interface doc (and the same for the other subroutines that are made private in other modules).

columnphysics/icepack_tracers.F90
```diff
@@ -907,7 +907,7 @@ end subroutine icepack_query_tracer_indices
 
       subroutine icepack_write_tracer_indices(iounit)
 
-        integer, intent(in), optional :: iounit
```
phil-blain, line 910: here `icepack_write_tracer_indices` is a public subroutine so we should re-run the autodocumentation script to update its interface in the doc to reflect the change of `iounit` from optional to mandatory.

phil-blain (commented):
Hi Tony,

Thanks a lot for these cleanups, these are very nice and make a lot of sense (and it's also safer, see the commit message of the commit I link to below). I left a few notes/questions below but mostly this is a direction I'm happy to see the code move towards.

Just a comment on the commit message, you mention:
> Clean up some intent statements

If I did not read too fast, that is in `columnphysics/icepack_therm_bl99.F90` and you are just moving the `intent` attributes on the same line as the argument they refer to, the actual intent (in, out or inout) does not change. I would make that clear in the commit message.

columnphysics/icepack_fsd.F90
```diff
@@ -230,7 +222,8 @@ subroutine icepack_init_fsd_bounds(nfsd, &
          c_fsd_range(n)=c_fsd1//'m < fsd Cat '//c_nf//' < '//c_fsd2//'m'
       enddo
 
-      if (l_write_diags) then
+      if (present(write_diags)) then
+      if (write_diags) then
```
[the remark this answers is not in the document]
apcraig: You are correct, thanks for noticing this @phil-blain.  I missed that and assume it behaved like everything else in the code, didn't look closely enough.  Overall, the diagnostics should either be on all the time OR they should be turned on via the interface.  It doesn't make sense to me that they be on by default but can be turned OFF by the interface.  That's counter to how everything works in CICE/Icepack in general.  What do others think?  For now, I think I'll update the call to icepack_init_fsd_bounds and set the flag to true.

columnphysics/icepack_shortwave.F90
apcraig: Only the public interfaces in icepack_intfc are auto-documented.  I'll confirm that and also rerun the documentation.

columnphysics/icepack_therm_vertical.F90
apcraig: Actually, this is odd.  In both the prior and current implementation, 

```
l_meltsliqn = c0
if (present(meltsliqn)) l_meltsliqn = meltsliqn
```
then later (line 2555 above)

`l_meltsliqn(n) = c0 
`

so I see no reason to copy meltsliqn into l_meltsliqn if it's ultimately intialized to c0.  The question is whether resetting l_meltsliqn(n) to c0 is a bug AFTER copying in meltsliqn.  Either l_meltsliqn is initialized c0 or meltsliqn.  Right now, it's confusing.  @eclare108213, any thoughts?  I may have messed this up in refactoring at some point, we should clarify and clean up.

njeffery: @apcraig : Looks to me like this first passing of meltsliqn to l_meltsliqn is not needed.  The term l_meltsliqn should be zeroed out once before thermo_vertical. Line 2555 is where it's done in mpas

apcraig: Thanks @njeffery, I'll clean that up.

apcraig:
I think I've updated the changes requested.  I retested full test suite on cheyenne with Icepack and CICE, everything passes and is bit-for-bit.  This is ready for another review/merge.

columnphysics/icepack_orbital.F90
eclare108213: This might complicate coupling in E3SM, if it's using CESMCOUPLED (I'm not sure).  @njeffery

columnphysics/icepack_shortwave.F90
eclare108213: Should `fswthru_vdr` and friends be optional in `absorbed_solar`?

eclare108213 (commented):
Nice cleanup, thanks.  Just a couple of comments below, and I'm flagging @njeffery and @darincomeau because there are changes here to icepack_mechred, snow and orbital that would need to be considered when coupling into E3SM.

columnphysics/icepack_shortwave.F90
apcraig: absorbed_solar is a local routine, it's not public.  Also, it operates on 1 category while the fswthru_vdr optional arguments are arrays of size ncat.  So we cannot actually pass the full optional argument thru.  So this means we have to create a local copy for that scalar and copy in/out of the optional arguments.  We could do various things to change this like having absorbed solar operate on all categories at once.  But for now, I just left the basic implementation as it was and adapted.  Those fswthru_vdr are always passed as currently implemented so adding optional wouldn't have an impact.  What we'd need to do in that case is add a new flag to say whether to compute those or not based upon the arguments passed into shortwave_ccsm3.  That won't affect results or performance.  If absorbed_solar becomes public, then we may have to rethink this.

columnphysics/icepack_orbital.F90
apcraig: Right, there are very few CESMCOUPLED cpps left in Icepack, we tried to clear them out a while ago and shift more to namelist settings.  Use of CESMCOUPLED in icepack_orbital was discussed of that at the time.  If this causes E3SM a problem, we can refactor it.  The problem is that it will probably require a change in all the models using CESMCOUPLED in terms of how the orbital stuff is handled, so we're trying not to change that until we have to.  But maybe we should do this at some point with consultation with CESM, RASM, etc so it's well coordinated.  New implementation still TBD as well.

eclare108213 (approved):
This looks good to me.  I'll approve now (because I'll be out the rest of the week) but I'd like @njeffery to take a look also.  @phil-blain does this satisfy your comments?

phil-blain (approved):
Looks good to me!

PR #431, opened 2023-03-15
Fix ReadTheDocs, update documentation

apcraig:
## PR checklist
    Fix ReadTheDocs build, update documentation
- [X] Developer(s): 
    apcraig
    no testing done, just doc changes
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,4 @@ Head over to our [Contributing][contributing] guide to learn more about how you 
 
 ## License
 See our [License](LICENSE.pdf) and [Distribution Policy](DistributionPolicy.pdf).
+
--- a/doc/source/developer_guide/dg_col_phys.rst
+++ b/doc/source/developer_guide/dg_col_phys.rst
@@ -84,7 +84,7 @@ Overall, columnphysics changes in the Icepack model should include the following
 
   * All modules should have the following set at the top
 
-    .. code-block:: fortran
+    .. code-block::
 
        implicit none
        private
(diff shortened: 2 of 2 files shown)


apcraig:
This PR is trying to debug the new readthedocs errors.

apcraig:
Killing this, need to debug another way.

apcraig:
It's not the badge, that's the first thing I checked.  I have created a new PR.  Had to go thru the readthedocs logs carefully and do some googling.  Seems to have been 2 minus signs that were "encoded" wrong.  You can see the diff in the new PR, #432.  But you can't tell there is a diff.  I deleted the two minus signs and typed them back in.  May have come from a copy and paste, hard to know.

If you look at the old documentation that had this, the html looks fine.  The pdf was generated but these two minus signs were missing the pdf documentation.

Apparently, readthedocs changed behavior in the last day or two and it now fails the documentation if a pdf error occurs, wasn't doing that before.

The part I don't like is that I can create a PR to Icepack without the fixes and the github actions readthedocs check passes.  So maybe that's just doing the html, not sure, but that could be a separate problem that we need to fix.  I need to try to understand what github actions is doing and why it's not catching the pdf issue.  May defer that, see how things go for a while.

apcraig:
I think the pdfs are nice to have.  Certainly easier to have a pdf on your laptop offline than to try to download the html directory or not have access when offline.  My feeling at this point is to see how this pans out.  If the pdf causes us constant problems, then lets create an issue and discuss.  In some ways, there was an error in the documentation (for latex anyway) and now we'll be trapping those more consistently, so that's good, I think.

Icepack docs badge is fixed with merge of #432.  Will do the same for CICE next.

PR #433, opened 2023-03-21
Fixes ice area inconsistency under convergence in single column mode

davidclemenssewall:

## PR checklist
    Changed Icepack such that when there is a closing forcing, ice with uniform properties is advected into the cell (so that closing does not create open water).
- [x] Developer(s): 
    David Clemens-Sewall
- Tony Craig, Elizabeth Hunke, Dave Bailey
    Icepack base_suite passed 138/138 (including new checks toggling 'ice_advection_type') in ubuntu docker container. CICE quick_suite passed 16/16 on cheyenne.
    - [ ] bit for bit
    - [x] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

Problem:
In the consortium main standalone configuration of Icepack, we read in a mechanical forcing file that contains and opening and a closing rate at each time step. When the forcing is net closing, the ridging scheme in Icepack produces open water (which is not physical) as a consequence of maintaining a constant cell area. In reality, and in CICE, this does not happen because it is ice advection from the surrounding area/grid cells that is driving the closing.

Proposed solution:
I've resolved this erroneous open water creation by adding an ice advection step prior to the ridging step (in the same way that CICE calls step_dyn_horiz before calling step_dyn_ridge). I assume that our single column grid cell is surrounded by ice of identical properties, so this amounts to scaling up the ice area, volume, and tracers before ridging. I have set the area of ice that is advected in to be equal to the closing rate times the timestep. I'd be happy to hear thoughts if others think a different area of ice should be advected in.

I've added a namelist parameter 'ice_advc_type' to icepack_in where 'uniform' means apply this advection step and 'none' means do nothing (a.k.a. the behavior without this fix). I've set the default to be 'uniform' because I think that most people running a single column model would not expect closing to produce open water.

When running standalone cases with the SHEBA ice dynamics forcing, including the advection of ice into the cell on closing dramatically increases the ice thickness and aice. This is to be expected since we are adding ice to the cell on closing. Note, ice is always advected out of the cell on opening (both in the current version and this modified version).


configuration/driver/icedrv_forcing.F90
```diff
@@ -454,6 +455,8 @@ subroutine get_forcing(timestep)
 
         opening(:) =   c1intp * open_data(mlast) + c2intp * open_data(mnext)
         closing(:) = -(c1intp * clos_data(mlast) + c2intp * clos_data(mnext))
+        !opening(:) = open_data(timestep)
+        !closing(:) = -1*clos_data(timestep)
```
eclare108213, line 459: If these lines are just for your local testing, please remove them.

configuration/driver/icedrv_RunMod.F90
```diff
@@ -177,13 +177,18 @@ subroutine ice_step
...
-        ! ridging
-        call step_dyn_ridge (dt_dyn, ndtd)
-
-        ! clean up, update tendency diagnostics
-        offset = c0
-        call update_state (dt_dyn, daidtd, dvidtd, dagedtd, offset)
+         
+         ! horizontal advection of ice into the single column
+         ! needed if we have closing (otherwise closing produces
+         ! open water)
```
eclare108213, line 183: Please make the comment a little more terse, e.g. 

`! horizontal advection of ice or open water into the single column`

without the additional explanation (which you've added to the docs, thank you)

configuration/driver/icedrv_step.F90
```diff
@@ -698,6 +700,85 @@ subroutine step_dyn_wave (dt)
...
+                                                   aicen=aicen(i,:),                 &
+                                                   vicen=vicen(i,:),                 &
+                                                   vsnon=vsnon(i,:),                 &
+                                                   aice0=aice0(i),                   &
+                                                   closing=closing(i) )               
+         
+                  endif ! tmask
+         
+                  enddo ! i
+               elseif (trim(ice_advc_type) /= "none") then
```
eclare108213, line 770: To make this section of the code more clear, I suggest 

`   elseif (trim(ice_advc_type) == "none") then `
`     ! do nothing:  no ice will be advected in, thus increasing the open water area through ridging`
`  else`
`     [abort] `

but it's okay like this.

configuration/driver/icedrv_step.F90
```diff
@@ -717,6 +798,7 @@ subroutine step_dyn_ridge (dt, ndtd)
       use icedrv_state, only: trcrn, vsnon, aicen, vicen
       use icedrv_state, only: aice, aice0, trcr_depend, n_trcr_strata
       use icedrv_state, only: trcr_base, nt_strata
+      use icedrv_state, only: divu
```
eclare108213, line 801: is divu used?

eclare108213 (changes requested):
I changed the title of this PR because ridging can create open water in exactly the manner you outline here, when neighboring grid cells have less ice area in them than the central cell.  It is also understandable that that behavior might be unexpected and undesirable in stand-alone Icepack tests. I like what you've done here, providing 2 cases with either no ice in the (imaginary) neighboring cells or with the same distribution of ice across them.  This is a nice enhancement for Icepack users, and it should not impact CICE or other drivers at all.  Thank you for contributing it!

The subroutine icepack_step_advection_scm only has a dozen or so lines of "real" code - the rest of it is checking conservation.  Since this option should never be used in a full sea ice model, the code for it could/should be kept in the driver rather than the column physics.  The dozen or so lines of active code can be inlined in step_advection_scm.  This would be more like CICE currently, where advection is not included in the column physics at all.

Tracer conservation should happen automatically due to the manner in which the tracers are defined (as densities, not absolute amounts).  Once you defined the changing areas properly, was that the case?  It's good that you've checked conservation, and now that you're satisfied (?), all of that code can be removed. (It will still be available in your git commit.)

I suggest changing the namelist flag value from 'uniform' to 'uniform_ice' throughout.  That's me being picky...

Please also add a couple of tests to Icepack's base_suite, which toggle ice_advc_type.

Finally, we've been moving toward less cryptic option names, so maybe `ice_advection_type`?

eclare108213:
Also, please, would you fix the formatting in the template above?  Checkboxes need to be
>     - [x]
instead of 
>     - [X ] 

Thank you!

configuration/driver/icedrv_step.F90
```diff
@@ -732,7 +814,12 @@ subroutine step_dyn_ridge (dt, ndtd)
       integer (kind=int_kind) :: &
          i,            & ! horizontal indices
          ntrcr,        & !
-         nbtrcr          !
+         nbtrcr,       & !
+         j               ! ice category index
+      
+      real (kind=dbl_kind) :: &
+         closing_area, & ! fractional area closed due to ice dynamics
+         aice_post_advection ! ice area after advecting ice to fill in closed area
```
eclare108213, line 822: Are these variables used?

apcraig:
Had a quick look, we need to be thoughtful about adding new public interfaces to Icepack's columnphysics.  If icepack_step_advection_scm can be migrated out to the driver, then I support that effort.  I assume that will require some additional queries for columnphysics "data", and is why it was put inside the columnphysics in the first place.  Hopefully, all the data that is needed is accessible to the driver.  Let me know if there are any questions.

configuration/driver/icedrv_RunMod.F90
davidclemenssewall: done

configuration/driver/icedrv_step.F90
davidclemenssewall: done

configuration/driver/icedrv_step.F90
davidclemenssewall: No. I've removed the line. Here's a long-winded explanation of what I was thinking about:

I had originally included divu here because divu (and shear for that matter) are always equal to zero. I thought that icepack_step_ridge was using the rdg_conv and rdg_shear values to calculate opening and closing. Then I would update the divu value as `divu(i) = opening(i) - closing(i)` after the call to icepack_step_ridge. In standalone icepack runs that's not the case because we simply read in the opening and closing. Also, if the ridging step needs to iterate (as it does anytime we have closing and `ice_advection_type = "none"`) opening and closing get overwritten anyways. We could compute divu from opening and closing before running the ridging step, but that could also be computed directly from the forcing data. So in the current configuration, I'm not sure it makes sense to include divu and shear in the history output at all. I guess it saves a user from having to query the forcing files in post-processing.

Also, I'm chatting with the ice dynamics experts from MOSAiC to try to determine how to relate the observations we've made to an appropriate dynamics forcing for single column models. I'm not sure yet if our resulting product will be opening and closing rates or something else (e.g., like observations of the net advection and amount of ridging convergence and shear). So my plan is to revisit the whole divu question after we sort out the form of the dynamics forcing.

configuration/driver/icedrv_step.F90
davidclemenssewall: No, they were for testing. I've removed them.

configuration/driver/icedrv_forcing.F90
davidclemenssewall: done.

configuration/driver/icedrv_step.F90
```diff
@@ -698,6 +700,85 @@ subroutine step_dyn_wave (dt)
...
+            if (icepack_warnings_aborted()) call icedrv_system_abort(string=subname, &
+                file=__FILE__,line= __LINE__)
+   
+         !-----------------------------------------------------------------
+         ! Ice advection
+         !-----------------------------------------------------------------
+            ! Currently we only do ridging for the SHEBA ocean data type (in step_dyn_ridge)
+            if (trim(ocn_data_type) == "SHEBA") then
+               ! Currently only uniform (and none) advection is implemented
+               if (trim(ice_advc_type) == "uniform") then
```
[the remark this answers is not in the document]
davidclemenssewall: done.

doc/source/icepack_index.rst
```diff
@@ -224,6 +224,7 @@ either Celsius or Kelvin units).  Deprecated parameters are listed at the end.
    "Hstar", "determines mean thickness of ridged ice", "25. m"
    "**I**", "", ""
    "i0vis","fraction of penetrating visible solar radiation", "0.70"
+   "ice_advc_type", ":math:`\bullet` how to handle ice advection due to closing in single column mode (uniform = advect uniform ice, none = advect open water)", "uniform"
```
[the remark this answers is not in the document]
davidclemenssewall: done.

configuration/scripts/icepack_in
```diff
@@ -121,6 +121,7 @@
     atm_data_type   = 'clim'
     ocn_data_type   = 'SHEBA'
     bgc_data_type   = 'default'
+    ice_advc_type   = 'uniform'
```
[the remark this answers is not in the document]
davidclemenssewall: I agree that 'uniform_ice' is better and I've changed it throughout. 

I can think of one particular context where someone *might* want the ability to advect ice in based on input data. On field campaigns (including MOSAiC) for logistical reasons we often do not establish all of our mass balance measurement sites simultaneously. For example, mass balance sites at MOSAiC were installed over a 2.5 month period from late October to early January. There are various ways that one could handle using these observations in the context of single column modeling, one of which involves prescribing ice advection. Essentially, when a new measurement site is added advect in ice of a fractional area that the new site comprises with the properties of the new site while simultaneously advecting out that same fractional area of the grid cell (so that we don't create ridging).

I don't know if anyone is likely to implement something like that anytime soon (or ever) but I think naming this option 'uniform_ice' won't conflict if so.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -345,6 +345,8 @@ forcing_nml
    "``formdrag``", "logical", "calculate form drag", "``.false.``"
    "``fyear_init``", "integer", "first year of atmospheric forcing data", "1998"
    "``highfreq``", "logical", "high-frequency atmo coupling", "``.false.``"
+   "``ice_advc_type``", "``uniform``", "advect in uniform ice when closing in single column mode", ""
```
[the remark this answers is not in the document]
davidclemenssewall: done.

doc/source/user_guide/ug_running.rst
```diff
@@ -747,6 +747,15 @@ cases.  Current filenames can be found in the options scripts in
    about 4 :math:`W/m^2` for biases in the forcings or the model. In the namelist, set ``atm_data_type = clim`` 
    to use climatological atmospheric forcing.
 
+Horizontal ice advection
+------------------------
+
+When Icepack is run in standalone (i.e., single column) mode with a dynamical forcing (e.g., ``ocn_data_type = SHEBA``),
+closing implies the advection of area into the grid cell. The default assumption
```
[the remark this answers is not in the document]
davidclemenssewall: done.

davidclemenssewall:
@eclare108213 @apcraig @dabail10, this PR is ready for review.

I have inlined the changes in step_advection_scm and removed all modifications to the columnphysics. Because tracers are stored a densities it no longer modify the tracer values (essentially the scaling I was doing to the tracer values was just undone by the compute_tracers call). I've also made the requested documentation, naming and testing changes. The current commit passes the icepack base_suite and CICE quick_suite.

eclare108213 (commented):
The testing options are a little unclear.  You created a set_nml.advcopenw file which sets the new namelist option `ice_advection_type` to 'none'. There's not a file for 'uniform_ice';  you're initializing it to 'uniform_ice' (and also say that's the default in the doc). The _current_ default is 'none' (for base_suite baselines), so I would think that this new option would need to be set to 'none' to get the same results for all of the other tests in the base_suite.  What am I missing ?  I don't think this PR should change the results of all of the other tests.

davidclemenssewall:
I wonder if we should stick with option 1 until we have a mutually consistent forcing dataset. 

I've done a little bit of ad hoc analysis of the output and it definitely changes the ice state. However, since the default forcing doesn't correspond to a set of conditions that actually occurred anywhere, it's hard to evaluate whether the changes in the ice state are what we should expect. For example, with the defaults and no advection cells 2 and 3 (the two that start with ice in them in January) simulate more open water than SHEBA (peak around 30%, vs. 17% observed) and the timing is offset (SHEBA peak open water was in mid-August). Including ice advection the simulated maximum open water fraction is just 5% (for cells 2 and 3) but the timing is more consistent with the SHEBA observations. However, the ice thicknesses in the no advection case are more reasonable, and appear to be stable. In contrast, in the advection case the ice thickness is not stable but increases without limit if we run the simulation out several years. Which is not too surprising since we don't have any ice strength feedback (i.e., more ice is pushed into the grid cell and ridging happens even when the ice gets thick enough that it should resist further ridging). But I think this behavior might be unexpected for testing.

With the MOSAiC test case (and hopefully other future test cases with self-consistent atmosphere, dynamics, and ocean forcing) I think that it would make sense to turn on ice advection on closing as the default.

aice, vice, and vsno without advection (default settings):
no_adv

aice, vice, and vsno with advection (default settings otherwise):
clos_adv_inline

eclare108213:
Thank you, this is very helpful.  I agree that sticking with option 1 makes the most sense for now, considering the ice thickness behavior.  In a simulation with horizontal extent, ice would be advected both in and out, while in this PR you're only adding ice to the grid cell when it's converging. Does it make sense to also remove some of it when diverging?  That might be getting pretty complicated (like the real advection scheme), but it's worth thinking about -- seems like this will be a persistent issue for MOSAiC simulations in a single column.

davidclemenssewall:
Hmm, I had thought that Icepack was advecting ice out when we had divergence. We certainly get a drop in aice when there is an opening forcing. Now that I go back and look at the code though, I'm not totally certain how that is happening. Line 1289 in icepack_mechred (subroutine ridge_shift) increases aice0 by opning*dt. If there were no closing forcing, I think that this would be the only change in the ice thickness distribution at this step. However, then when asum_ridging is called (line 410) asum should be greater than 1. Which would cause iterate_ridging to be true and closing_net should then be equal to the opening rate from the first iteration. On the next iteration of ridge_shift I would think this would just close the open water that was opened on the first step. So I'm not sure how it is that Icepack is producing open water at all.

Maybe in step_advection_scm we should set `expansion_ratio = c1 + (closing(i) - opening(i)) * dt`?
That would at least be consistent with the assumptions that ridge_ice currently makes in line 299 (in icepack_mechred).

eclare108213:
The code distinguishes between advection, which moves ice between grid cells, and open water closing/opening within a grid cell via ridging.  Advection is necessarily a grid-dependent process and so is not included in Icepack's column physics, but the rates of divergence/convergence that result from the advection are used in the column physics to ridge ice in a single grid cell. And as ice ridges, it can cause the area of open water to increase -- that's what it should do, in the absence of other ice being advected in.  The expert on this part of the sea ice model is there at NCAR: @whlipscomb.  Maybe the two of you can put your heads together and propose a solution for Icepack with the MOSAiC data.  This is definitely one of the challenges of working with Icepack as a Lagrangian model!  Thanks for digging in.

davidclemenssewall:
@whlipscomb and I had a good discussion about this yesterday.

In short, our conclusion was that for the mechanical redistribution in single-column Icepack to be consistent with the mechanical redistribution of Icepack inside CICE, there needs to be an "advection" step that changes the total grid cell area by an amount consistent with the net opening and closing.

Without this step, Icepack does not handle closing **or opening** consistently with how CICE handles it. We've discussed the issues with closing at some length, but the opening issues are a new realization. In effect, what is happening in the current default configuration of Icepack is that when we provide it an opening forcing, it accomplishes the required amount of opening by ridging ice within the grid cell, not by advecting ice out of the grid cell. In other words, opening and closing have exactly the same effect in the current default version of Icepack. You can see in the following figure where we have opening followed by a similar magnitude of closing. Both the opening and closing decrease aice by a similar amount and neither has any impact on the trajectory of vice.
march_adv_none

By adding including an advection step before this (where we have `expansion_ratio = c1 + (closing(i) - opening(i)) * dt`) we get more intuitive results where opening decreases the ice area and volume and closing increases both:
march_adv_net

I think that the gradual decline in ice area with opening followed by an abrupt rise is due to the mixed layer response. The mixed layer starts a little above the freezing temperature and it takes a few hours of opening before enough heat is extracted to drop sst to the freezing temperature. Once it cools enough, frazil ice is rapidly added reducing the open water area.

With applying advection for opening and closing (the current change), the system appears stable and produces reasonable snow and ice thicknesses (see 5 year run below). Additionally, the amount and timing of open water is now much more consistent with SHEBA (where the dynamic forcing comes) than the default case. In the simulation the maximum open water fraction 15% on August 5. At SHEBA the max was 18% on August 7. SHEBA flights only measured open water every 2 weeks. The atmospheric and oceanic forcing is different anyways so I think this is a helpful gut check but nothing more.
fiveyear_adv_net

Because the current version does not handle opening consistently with CICE, I think we need to change the default and do a full comparison of all tests. I am happy to have the default either be assuming the simulated grid cell is surrounded by uniform ice of the same properties or by open water (but now actually advecting in the open water or uniform ice and advecting out ice on opening to be consistent with CICE). I have a slight preference for the default to be uniform ice but I'm happy either way. @eclare108213 @apcraig @dabail10 let me know what you think and I will implement it and do the full test comparison.

eclare108213:
Nice, I like what you've done here.  My vote would be for assuming that the grid cell is surrounded by ice that is the same as in the central cell, so that you aren't imposing some sort of spatial dependence on your results, and then have a test or two for the 'none' option (as you do now), for code coverage purposes.

davidclemenssewall:
Apologies for my delay here, we had some family stuff come up.

I have finished the changes and compared the output from all of the tests before and after this fix. None of the state variables in the test output changes dramatically or unreasonably after the fix (e.g., ice thickness changes by 10s of cm). So I don't think it is likely to disrupt other testing.

In general, the largest impact is on aice, as discussed above. There are also downstream impacts. Specifically, because the change reduces the open water fraction in the summer, there is less solar heating of the mixed layer and less basal melt. Also, having higher ice area in September increases the snow depth as less snowfall is lost to the ocean.

Here are plots of output for all tests without the fix:
[plots_main_20230421.zip](https://github.com/CICE-Consortium/Icepack/files/11301942/plots_main_20230421.zip)

And here with the fix:
[plots_convergence_open_water_6.zip](https://github.com/CICE-Consortium/Icepack/files/11301943/plots_convergence_open_water_6.zip)

configuration/driver/icedrv_step.F90
```diff
@@ -764,7 +764,34 @@ subroutine step_advection_scm (dt)
...
+                        ! below 1 and then in the second ridging iteration loop
+                        ! 'opning' will be set such that it adds enough open water
+                        ! to return the cell area to 1.
+                        ! If the forcing is net opening, we still need to advect
+                        ! ice out of the grid cell as above.
+                        expansion_ratio = c1 + (closing(i) - opening(i)) * dt
+                        if (expansion_ratio < 1) then ! net opening
+                           aice0(i) = aice0(i) * expansion_ratio
+                           do n = 1, ncat
+                              ! Advect ice out of cell
```
eclare108213, line 786: This comment could be a little confusing to readers, since ``ice_advection_type == 'none'``.  Maybe just say "remove ice from cell"?

doc/source/user_guide/ug_running.rst
```diff
@@ -747,6 +747,15 @@ cases.  Current filenames can be found in the options scripts in
    about 4 :math:`W/m^2` for biases in the forcings or the model. In the namelist, set ``atm_data_type = clim`` 
    to use climatological atmospheric forcing.
 
+Horizontal ice advection
```
eclare108213, line 750: I think that referring to any of these changes as "ice advection" is going to be confusing, since advection isn't technically done in the column physics, and there's no attempt at volume conservation, which _should_ always be conserved under advection.  Maybe the ``ice_advection_type`` option could be renamed something like ``surrounding_ice_type``?  And then make it explicit in the documentation that this allows the effects of advective processes to be represented in a single grid cell; in Icepack the uniform ice option is not conservative while the 'none' option is. ?

configuration/driver/icedrv_step.F90
```diff
@@ -732,8 +838,8 @@ subroutine step_dyn_ridge (dt, ndtd)
       integer (kind=int_kind) :: &
          i,            & ! horizontal indices
          ntrcr,        & !
-         nbtrcr          !
-
+         nbtrcr       
+
```
eclare108213, line 842: please remove superfluous white space

configuration/driver/icedrv_step.F90
```diff
@@ -698,6 +700,110 @@ subroutine step_dyn_wave (dt)
...
       end subroutine step_dyn_wave
 
+
+!=======================================================================
+!
+! Run one time step of horizontal advection (if there is closing)
+!
+! authors: David Clemens-Sewall, NCAR
+
+      subroutine step_advection_scm (dt)
```
eclare108213, line 710: See comments re nomenclature elsewhere

doc/source/user_guide/ug_running.rst
davidclemenssewall: I agree we don't want to confuse anyone. What if we renamed `ice_advection_type` to be `lateral_flux_type` and the choices could be either 'uniform_ice' or 'open_water' (with the default still being 'uniform_ice')? I think that makes it more clear that ice/open water is entering/exiting the grid cell. And, at least to me, 'flux' doesn't carry the same volume conservation connotation as 'advection'. Also, note that the 'none'/'open_water' option is only conservative when the forcing is net convergence. When the forcing is net divergence, we must remove ice volume from the single grid cell (i.e., the open water produced by a net divergence should reduce the mean thickness in the grid cell).

Below is a proposal for how to change the documentation to make this clear. Please let me know if this makes sense. Thanks!

Horizontal ice advection
------------------------

When Icepack is run in standalone (i.e., single column) mode with a dynamical forcing (e.g., ``ocn_data_type = SHEBA``),
closing implies the lateral flux of ice or open water area into the grid cell. The default assumption 
(in the namelist, ``lateral_flux_type = 'uniform_ice'``) is that the simulated grid cell is surrounded by grid cells with identical
ice properties to the simulated grid cell. When the dynamical forcing is net convergence, this uniform ice is fluxed into the
grid cell. Alternatively, one may assume that the simulated grid cell is surrounded by open water (in the namelist
``lateral_flux_type = 'open_water'``). Note that assuming the grid cell is surrounded by open water means that closing (i.ice convergence) will produce open water in the grid cell. In either case, when the forcing is net divergence, ice area
and volume are removed from the grid cell to accommodate the formation of open water implied by the net divergence.

configuration/driver/icedrv_step.F90
davidclemenssewall: Do you think it would make sense to rename `step_advection_scm` to `step_lateral_flux_scm`?

doc/source/user_guide/ug_running.rst
eclare108213: I like your suggestion to use `lateral_flux_type` instead.  Here's how I would do the documentation:

Horizontal ice advection
------------------------

When Icepack is run in standalone mode with a dynamical forcing (e.g., ``ocn_data_type = SHEBA``),
closing implies the lateral flux of ice or open water area into the grid cell. The default assumption 
(in the namelist, ``lateral_flux_type = 'uniform_ice'``) is that the active grid cell is surrounded by grid cells with
identical ice properties to the active grid cell, i.e. the ice is uniform across all of those cells, and when the dynamical forcing is net convergence, this uniform ice is 
fluxed into the grid cell. Alternatively, one may assume that the active grid cell is surrounded by open water (in the 
namelist ``lateral_flux_type = 'open_water'``), in which case 
closing (i.e., ice convergence) will produce open water in the grid cell. In either case, when the forcing is net 
divergence, ice area and volume are removed from the grid cell to accommodate the formation of open water implied 
by the net divergence.

configuration/driver/icedrv_step.F90
davidclemenssewall: Done.

davidclemenssewall:
I'm unsure why Read the Docs is now failing. I don't think that I have changed anything related to that. Otherwise I think this should be all set to review. It passes 138/138 on the base_suite.

apcraig:
Seems to be an issue with readthedocs.  Can you make a trivial change to your branch and push the commit.  Lets trigger the checks again and see if the problem persists.

apcraig:
OK, failed again.  I then forced a rebuild and it worked.  I think it's readthedocs at the moment.  At any rate, there is an updated readthedocs for this PR, so I think we're OK for the moment.  Let me know if you notice anything funny again.  Thanks.

davidclemenssewall:
Just checking in here, is there anything more that you'd like me to do on this PR? Otherwise, I think it is ready.

apcraig:
Just to be clear, are all results in Icepack and CICE expected to change?  Has anyone done a QC test in CICE?  I can run a full CICE test suite on cheyenne and do a QC test if it's needed/helpful.  Just let me know.

Quick review on my part shows no problems.  There is an empty line with just spaces added in icepack_mechred, but I can fix that later.  I have an automated tool that I should run on the code again.  It's also likely readthedocs is broken, but not due to this PR.  That's another thing we'll need to fix asap.  For now, I think once we feel testing is complete and @eclare108213 gives approval, we can probably merge.

apcraig:
Looked at the code again, this does not change anything in the Icepack columnphysics, so cannot affect CICE.  Is that what we want/expect?

eclare108213:
> Looked at the code again, this does not change anything in the Icepack columnphysics, so cannot affect CICE. Is that what we want/expect?

Yes, this only changes how Icepack is forced, so shouldn't affect CICE at all.

eclare108213 (approved):

apcraig (approved):

PR #411, opened 2022-11-11
finalize 0-layer thermo and cesm ponds deprecation in Icepack

eclare108213:
- [x] Short (1 sentence) summary of your PR: 
Removes all ifdef options allowing 'undeprecation' of 0-layer thermo and cesm meltponds, and updates documentation
- [x] Developer(s): 
@eclare108213 @dabail10 
- [x] Suggest PR reviewers from list in the column to the right.
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
    ENTER INFORMATION HERE 
- How much do the PR code changes differ from the unmodified code? 
    - [x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on CICE or any other models?
    - [x] Yes
    - [ ] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/.)
    - [x] Yes
    - [ ] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [ ] No 
- [ ] Please provide any additional information or relevant details below:

PR #448, opened 2023-07-29
Deprecate zsalinity

### LINKED ISSUE #439: Deprecate zsalinity
I'm in the process of deprecating zsalinity.  It looks like the "UNDEPRECATE" implementation done earlier just made it impossible to run with zsalinity, but there is lots of code that actually has to still be removed.  Is this correct?  Should I try to do that or is this better done by someone with a better understanding of zsalinity?

apcraig:
## PR checklist
    Deprecate zsalinity
- [X] Developer(s): 
    apcraig
    Icepack test suite is bit-for-bit on cheyenne with intel, gnu, and pgi EXCEPT for pgi with bgc (this must be a compiler optimization issue with the code changes).  https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#13e1c1dc418c7c38dedb1b9e76b6f72e53dd801a.  Final CICE with zsalinity deprecation testing is underway, but expected to be bit-for-bit in full test suite on cheyenne with intel, gnu, pgi EXCEPT pgi with bgc on (passes but not bit-for-bit).  Results for CICE for gnu and pgi are here, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#f836c7d5b39f7c3dc272ab50f6db615341110343
    - [X] bit for bit except PGI with bgc turned on
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

- Closes #439 
- Remove basic zsalinity code, but maintain Icepack interfaces for backwards compatibility.  If solve_zsal is set to true, Icepack will abort.
- Includes removing "bgc_S" variables and implementation
- Update documentation
- Will PR CICE changes and update Icepack in CICE once this PR is merged.
- Also tested this version of Icepack with an unmodified version of CICE to confirm backwards compatibility.  CICE will build and run bit-for-bit with this version of Icepack with deprecated zsalinity.


eclare108213:
@njeffery  -  Can the brine height tracer be run without zsalinity, i.e. with mushy thermo?  My plan has been to deprecate zsalinity but not hbrine, and I'm wondering whether that make sense.  Our tests only have them both turned on, not hbrine by itself.

njeffery:
Definitely.  The brine height tracer is needed for the biogeochemistry and doesn't depend on the specific halodynamics.


n


[1669832618850]

Nicole Jeffery

Scientist

Los Alamos National Lab

Computational Physics and Methods, CCS-2


mobile: (505) 257-8998

email: ***@***.***



From: Elizabeth Hunke ***@***.***>
Sent: Monday, July 31, 2023 12:49:50 PM
To: CICE-Consortium/Icepack
Cc: Jeffery, Nicole; Mention
Subject: [EXTERNAL] Re: [CICE-Consortium/Icepack] Deprecate zsalinity (PR #448)


@njeffery<https://urldefense.com/v3/__https://github.com/njeffery__;!!Bt8fGhp8LhKGRg!HRqXKyuOdSNJ4w1CrikcIJbdDqF_AVjlP4qKfoNsh9ZdpmGz9d3VViAl2v7TIUWopGmUD2BuBZpQpkDLV5MNw2zY_A$> - Can the brine height tracer be run without zsalinity, i.e. with mushy thermo? My plan has been to deprecate zsalinity but not hbrine, and I'm wondering whether that make sense. Our tests only have them both turned on, not hbrine by itself.

eclare108213:
Excellent.  Correction:  we do have tests with brine tracer on and zsalinity off in both icepack and cice.  They are both on only for the zsal test.

columnphysics/icepack_brine.F90
```diff
@@ -1000,7 +765,8 @@ end subroutine icepack_init_hbrine
 
 !=======================================================================
 !autodocument_start icepack_init_zsalinity
-!  Initialize zSalinity
+!  **DEPRECATED**, all code removed
+!  Interface provided for backwards compatibility
```
eclare108213, line 769: I'm not sure I understand what would break if this were removed.  Could the interface be put in an 'ifdef undeprecate' for the next release and then removed completely after that?  Actually, since we had partially deprecated salinity earlier, it was broken and so I think notice has been served -- no one should be calling it at this point.  Is the backwards compatibility issue just so that other (non-CICE) drivers can update icepack without having to make changes in the driver itself?

columnphysics/icepack_tracers.F90
```diff
@@ -96,7 +96,7 @@ module icepack_tracers
          nt_bgc_PON   = 0, & ! zooplankton and detritus
          nt_bgc_hum   = 0, & ! humic material
          nt_zbgc_frac = 0, & ! fraction of tracer in the mobile phase
-         nt_bgc_S     = 0    ! Bulk salinity in fraction ice with dynamic salinity (Bio grid)
+         nt_bgc_S     = 0    ! Bulk salinity in fraction ice with dynamic salinity (Bio grid) (deprecated)
```
eclare108213, line 99: why does nt_bgc_S need to be kept?

doc/source/user_guide/lg_sequence.rst
```diff
@@ -123,10 +123,10 @@ computation and a routine to update the radiation computation::
       use icepack_shortwave, only: icepack_prep_radiation
       use icepack_shortwave, only: icepack_step_radiation
 
-icepack_brine address brine and zsalinity computations::
+icepack_brine addresses brine computations::
 
       use icepack_brine, only: icepack_init_hbrine
-      use icepack_brine, only: icepack_init_zsalinity
+      use icepack_brine, only: icepack_init_zsalinity  ! DEPRECATED
```
eclare108213, line 129: why not remove this line?

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -436,7 +436,7 @@ zbgc_nml
    "``f_exude_l``", "real", "fraction of exudation to DOC lipids", "1.0"
    "``f_exude_s``", "real", "fraction of exudation to DOC saccharids", "1.0"
    "``grid_o``", "real", "z biology for bottom flux", "5.0"
-   "``grid_oS``", "real", "z salinity for bottom flux", "5.0"
+   "``grid_oS``", "real", "DEPRECATED", ""
```
eclare108213, line 439: In previous code deprecations we have moved lines like this into comments, e.g. see https://github.com/CICE-Consortium/Icepack/pull/411/files#diff-5e3296389949243ad10cff84e0cf3b0b81329d2c9391920fb4cd5b4367018d90

eclare108213 (commented):
Some questions -

apcraig:
Good questions about the deprecation.  I kept the zsalinity public interfaces and arguments for the time being for backwards compatibility.  They are not needed for the Icepack driver or a future version of CICE, but other users might need them.  If a user has an older version of CICE or uses Icepack in their ice model (MPASSI), then an upgrade to this version of Icepack should not break anything.  They can compile and run (with zsalinity off) without having to change any of their code.  If they turn on zsalinity, the code should abort.  If we remove the public interfaces and arguments, then an Icepack upgrade could/will immediately fail to build and the user will have to modify calls into Icepack to get things working again, even if they are not using zsalinity.

So we can take either approach.  Provide full backwards compatibility by maintaining zsalinity arguments/interfaces that have no impact on the code.  Or remove that code and possibly force all users to update their "driver" when they bring in a new version of Icepack, with no real benefit.  There are pros and cons to both approaches.

I tested the updated version of Icepack (the one in this PR) with the baseline version of CICE (with all the zsalinity code/calls) to make sure it truly is backwards compatible and bit-for-bit.  Of coarse, CICE only works in that mode if zsalinity is off.

apcraig:
If you feel we should fully deprecate the Icepack interfaces for zsalinity, I can certainly do that easily.

apcraig:
I think hbrine is tested with bgc.  Do we need to add a test to Icepack and/or CICE with just hbrine without bgc?  Is that possible?  If so, what other options should be turned on with it?

apcraig:
Is there anything else to do with the PR, is it ready to merge?  Please see also https://github.com/CICE-Consortium/CICE/pull/851, that also needs a review.

apcraig:
I just rebased this, should be ready for review and merge.

eclare108213 (approved):
This looks good to me.  I'd like to see more of the deprecated code disappear completely, but I agree that it's better to not break users' codes when they try to build.  Maybe we can get rid of the rest of it when we make a major release (which might break other things too - fix them all at once).  

My one concern is that we not break the brine tracer.  It looks like zbgc is tested in both icepack and cice with tr_brine=T, so this should be okay, especially since we'll be thoroughly testing the new BGC code as part of the E3SM/Icepack merge, and that will include the brine module.

apcraig:
I'm pretty sure the brine tracer is preserved with these changes.  The full test suite runs and is basically bit-for-bit.  Maybe we need to add a new brine tracer test at some point.

PR #459, opened 2023-09-26
DO NOT MERGE: Merge E3SM Icepack to Consortium

apcraig:


columnphysics/icepack_tracers.F90
```diff
@@ -1261,7 +1264,18 @@ subroutine icepack_compute_tracers (ntrcr,     trcr_depend,    &
                trcrn(it) = atrcrn(it) / aicen
             else
                trcrn(it) = c0
-               if (it == nt_Tsfc) trcrn(it) = Tocnfrz  ! surface temperature
+               if (it == nt_Tsfc) then
+! tcraig, these old options should be deprecated
+! exist for bit-for-bit backwards compatibility in testing
+                  if (tfrz_option == "mushy_old" .or. &
```
eclare108213, line 1270: This needs to be rectified between the codes.  It looks like they're only there to maintain BFB in Consortium tests.  Is it possible to do this merge in 2 PRs, first BFB and then non-BFB?

eclare108213 (commented):
There's a lot here, but not too bad.  There were only a handful of PRs directly into E3SM's fork, so we can grab documentation for the PR from those.  Looks like the snicar changes were added to the formal documenation but other things will still need to be added/updated.  I'd rather not add the _old options into main, or if we can do a BFB PR first, then do the non-BFB one quickly to get rid of them.  Do the snicar changes require updates to the forcing data? I don't think so.

apcraig:
I think the best process is to merge the PR more-or-less directly.  Then we can do a quick cleanup from that as a separate PR.  We can add documentation at either step.  I might suggest during cleanup, but can go either way.

eclare108213:
I prefer to update readthedocs with the PR where things are changing.  It's fine to merge in one big PR, but if it's not BFB I think we should run QC as a sanity check.  Then clean up.

apcraig:
Sounds good.  It should be BFB except for a few documented cases (modal aerosols in Icepack runs and non-ice values of TF on CICE restart files).  That should be under control.

If we want to update documentation with the PR, then we should do that on the existing branches (E3SM/Icepack and apcraig/CICE).  An alternative for the E3SM/Icepack branch would be for me to create a branch in my fork from that and we could PR from there after documentation changes, but I don't recommend that.  Do you want me to start working thru the documentation changes?  I thought we did the user guide / namelist kind of stuff as part of the original PRs to those branches.  Maybe there is some science updates to do?  The readthedocs documentation changes shouldn't be too much I hope?

I think the main thing is to document the individual commits in this PR well as part of the PR.  We will squash merge this to main and we want a nice record of what came in.  Again, I can start that process now if you like.  I just created this PR quickly so you'd have something to look at.

eclare108213:
Yes, please get started.  I think we updated readthedocs for the snicar changes, which is where the science updates would be.  Not sure about Tf - let's see what's already there and decide if more needs to be said, beyond adding the namelist options.  Hopefully not much is needed.

apcraig:
Will be opening a new PR.

PR #460, opened 2023-09-26
Merge E3SM Icepack cice-consortium/E3SM-icepack-initial-integration to Consortium

apcraig:
## PR checklist
    Merge E3SM Icepack cice-consortium/E3SM-icepack-initial-integration to Consortium
- [X] Developer(s): 
    apcraig, eclare108213, dabail19
    underway....
    - [x] bit for bit, except modal results
    - [ ] more substantial 
    - [X] Yes, will need to update CICE
    - [ ] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

- Update dEdd shortwave, snicar, https://github.com/E3SM-Project/Icepack/pull/7
  - Update snicar-ad table lookups, hardcode tables, provide additional options
  - Update dEdd implementation
  - Support dEdd 3-band and 5-band data
  - Support full snicar tables, hardcoded into model, requires USE_SNICARHC cpp
  - Support smaller test data table
  - Add icepack_shortwave_data file/module to store snicar table data
- Update modal aerosols, this changes answers in Icepack standalone
- Update public interfaces
   - remove single nspint parameter, add 3 and 5 band nspint parameters 
   - add icepack_init_radiation
   - add icepack_salinity_profile
   - add icepack_enthalpy mush
   - add icepack_warnings_getall
   - modify icepack_prep_radiation, remove scalar parameters (ncat, nilyr, nslyr)
   - modify icepack_step_radiation, remove scalar parameters (ncat, nblyr, nilyr, nslyr, dEdd_algae, calendar_type, days_per_year, nextsw_cday, modal_aero), remove snicar table data arguments
   - add Tf argument to icepack_compute_tracers, icepack_aggregate, icepack_step_ridge, 
 - Modify icepack parameters
   - add hi_min, minimum thickness allowed for thermo (real,m), https://github.com/E3SM-Project/Icepack/pull/10.  The current default could change answers in CESM CICE simulations, but the setting is now a CICE namelist option.
   - add Tliquidus_max, maximum liquidus temperature of mush (real, degC), https://github.com/E3SM-Project/Icepack/pull/13.  The default is 0.01 but it's also a new namelist in CICE.
   - add snw_ssp_table, lookup table options ('snicar', 'test')
   - add 'dEdd_snicar_ad' option to shortwave parameter
   - add 'constant' option to tfrz_option
- Update Tf implementation to make internally consistent in multiple places, https://github.com/E3SM-Project/Icepack/pull/12
   - remove hardcoded Tocnfrz in some places
   - add tfrz_option = 'constant' which sets Tf to Tocnfrz as an option
   - add "_old" options to tfrz_option for backwards compatibility (will be removed)
- Modify iDin computation in subroutine compute_microS_mushy
- Fix fhtan computation in compute_albedos subroutine
- Modify some internal subroutine interfaces, remove parameters from arguments, leverage module data instead
- Rename orb interfaces, change from shr_orb to icepack_orb to better differentiate internal and external versions of the orbital code
- Refactor some optional argument handling
- Clean up icepack driver namelist logging
- Clean up some formatting
- Add ability to reuse Icepack binary in Icepack test scripts
- Add snicar_suite test suite
- Update documentation

NOTE: TO DO AFTER MERGE

- Clean up "_old" tfrz_options and validate results
- Have USE_SNICARHC cpp be "on" all the time with option to turn off for testing
- Update CICE calls to pass frazil ice settings at initialization and remove frazil ice settings from runtime interfaces.  Then update Icepack to remove those optional frazil ice arguments in runtime interfaces.


apcraig:
I will review the documentation and propose some changes via a PR to [E3SM-Project:cice-consortium/E3SM-icepack-initial-integration](https://github.com/E3SM-Project/Icepack/tree/cice-consortium/E3SM-icepack-initial-integration).  Once that's ready, we can test/review/merge this PR to the Consortium.

columnphysics/icepack_itd.F90
```diff
@@ -27,15 +27,15 @@ module icepack_itd
 
       use icepack_kinds
       use icepack_parameters, only: c0, c1, c2, c3, c15, c25, c100, p1, p01, p001, p5, puny
-      use icepack_parameters, only: Lfresh, rhos, ice_ref_salinity, hs_min, cp_ice, Tocnfrz, rhoi
```
dabail10, line 30: Didn't we fix something like this in Icepack already?

dabail10 (approved):
If this is all bfb, then I don't see any issues. I was confused about the Tocnfrz versus Tf thing. I know we removed hard-coded -1.8 when possible, but then we decided not to update to Tf as this would change answers? Can someone remind me what happened here?

columnphysics/icepack_itd.F90
apcraig: No.  The Tf/Tocnfrz was implemented on the E3SM side, and we have not moved any of those changes over to the Consortium.  In hindsight, it probably should have been done on the Consortium side and moved to the E3SM side.  I think that was early days of the E3SM Icepack version, and we probably hadn't realized that we should do changes, when possible, in the Consortium.  That approach was adopted later.

eclare108213 (commented):
Just one comment here which is more appropriate for the documentation PR.  I'll add others there.

apcraig:
@dabail10, the Tf/Tocnfrz thing is a little confusing.  I think what was done was the following.  There were a few places where the freezing T was hardwired, those were modified so the code uses Tf (the computed value of freezing temp) and is internally consistent.  Then we added a new option for tfrz_option, "constant", which allows the user to set the value of Tocnfrz via icepack_parameters.  The default for Tocnfrz is -1.8.  The other options for tfrz_option is "minus1p8", "linear_salt", and "mushy".  The difference between "constant" and "minus1p8" is that the former allows the user to set a Tocnfrz value that will be used while the latter always uses -1.8.

We are also supporting "_old" options for all of these settings.  That provides backward compatibility for CICE testing to make sure we're not changing answers overall relative the Consortium main.  The "_old" version just sets trcrn(It) in icepack_compute_tracers to Tocnfrz (-1.8 by default) instead of Tf (the freezing temperature associated with the tfrz_option).  These "_old" versions will be deprecated after the code is moved to Consortium main, that will change answers but be more correct.  The "_old" versions exist just for testing.

apcraig:
Testing vs Icepack Consortium and CICE Consortium mains performed as expected.
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#96f2fc707fc743d7ce6eb8f543bd449a35a1a649
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#7603016e1db8cf6d752e3b35bfbe18391639d4db

- Icepack modal tests changed answers relative to Consortium main
- CICE restart files have different surface temp values at non-ice points, but log files are identical
- CICE boxchanl1e and boxchan1n changed answers, had to set tfrz_option to 'mushy_old' to recover prior results.

This is ready for review and merge.

eclare108213 (approved):
This code has been thoroughly tested in E3SM and is BFB in most cases for the Consortium.

PR #472, opened 2023-10-20
Update 5-band dEdd to support test data

apcraig:
## PR checklist
    Update 5-band dEdd to support test data
- [X] Developer(s): 
    apcraig
    Full test suite run on cheyenne with 3 compiler with Icepack and CICE, all results bit-for-bit.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#8916b9ff2c58a3a095235bb5b4ce7e8a68f76e87, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#7a0d4e9f16eb6505a11372ced01dbd217f00d505
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Add an interpolation method in icepack_shortwave.F90 for rsnw

Refactor portions of icepack_shortwave to improve robustness

Add snicar and snicartest test cases

Update the 5-band snicar shortwave rsnw_snicar_tab interpolation.  This changes answers but QC tests pass.  The new implementation checks whether the rsnw_snicar_tab is an array of integer values with deltas of value 1 (and sets the rsnw_snicar_chk variable).  If so, it uses a shortcut to find the rsnw_snicar_tab bounds for interpolation, otherwise it calls the shortwave_search routine.  In testing, the shortcut did not change performance much, but that could change in the future.

### DIFF
--- a/columnphysics/icepack_shortwave.F90
+++ b/columnphysics/icepack_shortwave.F90
@@ -62,7 +62,7 @@
       use icepack_tracers,    only: tr_zaero, nlt_chl_sw, nlt_zaero_sw
       use icepack_tracers,    only: n_algae, n_aero, n_zaero
       use icepack_tracers,    only: nmodal1, nmodal2, max_aero
-      use icepack_shortwave_data, only: nspint_3bd, nspint_5bd
+      use icepack_shortwave_data, only: nspint_3bd, nspint_5bd, rsnw_datatype
       use icepack_zbgc_shared,only: R_chl2N, F_abs_chl
       use icepack_zbgc_shared,only: remap_zbgc
       use icepack_orbital,    only: compute_coszen
@@ -4500,9 +4500,6 @@
          sza_factor   , & ! parameter for high sza adjustment
          mu0
 
-      character (len=char_len), save :: &
-         rsnw_snicar_chk = 'unknown'
-
       character(len=*),parameter :: subname='(compute_dEdd_5bd)'
 
 !-----------------------------------------------------------------------
@@ -4814,23 +4811,8 @@
                   ws = ss_alb_ice_drc     (ns,nmbrad_snicar)
                   gs = asm_prm_ice_drc    (ns,nmbrad_snicar)
                else
-                  ! check whether rsnw_snicar_chk are integers with delta of one, makes search faster
-                  if (rsnw_snicar_chk == 'unknown') then
-                     rsnw_snicar_chk = 'delta1'
-                     do n = 1,size(rsnw_snicar_tab)
-                        if (abs(rsnw_snicar_tab(n) - real(nint(rsnw_snicar_tab(n)),kind=dbl_kind)) > 1.0e-6) then
-                           rsnw_snicar_chk = 'mixed'
-                        endif
-                     enddo
-                     do n = 2,size(rsnw_snicar_tab)
-                        if (abs(rsnw_snicar_tab(n) - rsnw_snicar_tab(n-1) - c1) > 1.0e-6) then
-                           rsnw_snicar_chk = 'mixed'
-                        endif
-                     enddo
-!                     call icepack_warnings_add(subname//' rsnw_snicar_chk = '//trim(rsnw_snicar_chk))
-                  endif
                   ! linear interpolation
-                  if (trim(rsnw_snicar_chk) == 'delta1') then
+                  if (trim(rsnw_datatype) == 'sorted_idelta1') then
                      ! NOTE:  Assumes delta rsnw_snicar_tab is 1 and rsnw_snicar_tab are integers
                      ! This is just for performance, could call shortwave_search
                      nr = ceiling(rsnw(ksnow)) - nint(rsnw_snicar_min) + 1
@@ -4864,24 +4846,9 @@
                   ws = ss_alb_ice_dfs     (ns,nmbrad_snicar)
                   gs = asm_prm_ice_dfs    (ns,nmbrad_snicar)
                else
-                  ! check whether rsnw_snicar_chk are integers with delta of one, makes search faster
-                  if (rsnw_snicar_chk == 'unknown') then
-                     rsnw_snicar_chk = 'delta1'
-                     do n = 1,size(rsnw_snicar_tab)
-                        if (abs(rsnw_snicar_tab(n) - real(nint(rsnw_snicar_tab(n)),kind=dbl_kind)) > 1.0e-6) then
-                           rsnw_snicar_chk = 'mixed'
-                        endif
-                     enddo
-                     do n = 2,size(rsnw_snicar_tab)
-                        if (abs(rsnw_snicar_tab(n) - rsnw_snicar_tab(n-1) - c1) > 1.0e-6) then
-                           rsnw_snicar_chk = 'mixed'
-                        endif
-                     enddo
-!                     call icepack_warnings_add(subname//' rsnw_snicar_chk = '//trim(rsnw_snicar_chk))
-                  endif
                   ! linear interpolation
-                  if (trim(rsnw_snicar_chk) == 'delta1') then
-                     ! NOTE:  Assumes delta rsnw_snicar_tab is 1 and rsnw_snicar_tab are integers
+                  if (trim(rsnw_datatype) == 'sorted_idelta1') then
+                     ! NOTE:  delta rsnw_snicar_tab is 1 and rsnw_snicar_tab are integers
                      ! This is just for performance, could call shortwave_search
                      nr = ceiling(rsnw(ksnow)) - nint(rsnw_snicar_min) + 1
                   else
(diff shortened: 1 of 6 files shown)


eclare108213:
@proteanplanet I invite you to take a look at this PR.  @apcraig is setting up an Icepack test for the 5-band shortwave and needed to generalize the hardwired 5-band coding to do that.  Personally, I think we should fix the hard-wired weirdnesses in here (see the code diff) to just do the linear interpolation all the time, but that will change answers.  @apcraig has modified the code in a way that allows our testing without changing the 5-band shortwave answers. We can fix this later, but I'm pinging you here in case you'd like it done sooner.  I'm not very happy with the current 5-band coding.

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,22 +4810,38 @@ subroutine compute_dEdd_5bd(                          &
...
-                  delr = (rsnw(ksnow) - floor(rsnw(ksnow))) &   ! hardwired delta radius = 1 in table
-                       / (ceiling(rsnw(ksnow)) - floor(rsnw(ksnow))) ! denom always = 1
-                  ks = ext_cff_mss_ice_drc(ns,nr-1)*(c1-delr) &
-                     + ext_cff_mss_ice_drc(ns,nr  )*    delr
-                  ws = ss_alb_ice_drc     (ns,nr-1)*(c1-delr) &
-                     + ss_alb_ice_drc     (ns,nr  )*    delr
-                  gs = asm_prm_ice_drc    (ns,nr-1)*(c1-delr) &
-                     + asm_prm_ice_drc    (ns,nr  )*    delr
+               else
+                  if (trim(snw_ssp_table) == 'snicar') then
```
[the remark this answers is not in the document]
apcraig: correct.

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,22 +4810,38 @@ subroutine compute_dEdd_5bd(                          &
...
+                     ! NOTE:  Assumes delta rsnw is 1 and rsnw are basically integers
+                     nr = ceiling(rsnw(ksnow)) - nint(rsnw_snicar_min) + 1
+                     if (ceiling(rsnw(ksnow)) - rsnw(ksnow) < 1.0e-3_dbl_kind) then
```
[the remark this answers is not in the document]
apcraig: correct, don't know why the diff is just one side of rsnw(ksnow), don't know if it should be abs or not, don't know why 1.e-3 was chosen.

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,22 +4810,38 @@ subroutine compute_dEdd_5bd(                          &
...
+                        ks = ext_cff_mss_ice_drc(ns,nr)
+                        ws = ss_alb_ice_drc     (ns,nr)
+                        gs = asm_prm_ice_drc    (ns,nr)
+                     else
```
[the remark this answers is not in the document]
apcraig: correct.

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,22 +4810,38 @@ subroutine compute_dEdd_5bd(                          &
...
+                             / (ceiling(rsnw(ksnow)) - floor(rsnw(ksnow))) ! denom always = 1
+                        ks = ext_cff_mss_ice_drc(ns,nr-1)*(c1-delr) &
+                           + ext_cff_mss_ice_drc(ns,nr  )*    delr
+                        ws = ss_alb_ice_drc     (ns,nr-1)*(c1-delr) &
+                           + ss_alb_ice_drc     (ns,nr  )*    delr
+                        gs = asm_prm_ice_drc    (ns,nr-1)*(c1-delr) &
+                           + asm_prm_ice_drc    (ns,nr  )*    delr
+                     endif
+                  else
+                     ! linear interpolation
```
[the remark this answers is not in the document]
apcraig: My goal here was to leave the "snicar" snw table results bit-for-bit and add support for the "test" table.  That meant leaving all the logic and math as implemented.  Changing the logic or the math could/would change answers.  I agree that we could (should!) clean this up, but I propose it be done as a separate PR which would allow us to more carefully test the snicar results.  This PR, to some degree, was just to add support for the "test" table.

apcraig:
I  just updated the implementation to cleanup the standard snicar 5-band rsnw_snicar_tab interpolation.  This changes answers, but QC tests pass and the differences cannot be seen on the QC diff plot (attached).  I ran run a small test suite focused on the snicar changes and that all works as expected.  I am running a full test suite now, but don't expect any problems.

ice_thickness_cheyenne_intel_smoke_gx1_72x1_medium_qc_snicar qst1

ice_thickness_cheyenne_intel_smoke_gx1_72x1_medium_qc_snicar qsb1_minus_cheyenne_intel_smoke_gx1_72x1_medium_qc_snicar qst1

apcraig:
Full test suite on latest changes passes.

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,16 +4813,33 @@ subroutine compute_dEdd_5bd(                          &
...
+                  ! check whether rsnw_snicar_chk are integers with delta of one, makes search faster
+                  if (rsnw_snicar_chk == 'unknown') then
+                     rsnw_snicar_chk = 'delta1'
+                     do n = 1,size(rsnw_snicar_tab)
+                        if (abs(rsnw_snicar_tab(n) - real(nint(rsnw_snicar_tab(n)),kind=dbl_kind)) > 1.0e-6) then
+                           rsnw_snicar_chk = 'mixed'
+                        endif
+                     enddo
+                     do n = 2,size(rsnw_snicar_tab)
+                        if (abs(rsnw_snicar_tab(n) - rsnw_snicar_tab(n-1) - c1) > 1.0e-6) then
```
eclare108213, line 4826: why do you need 2 loops, one with n=1,size() using real(nint()) and the other with n=2,size() without real(nint()) but with -1?

columnphysics/icepack_shortwave.F90
```diff
@@ -4815,16 +4813,33 @@ subroutine compute_dEdd_5bd(                          &
...
-                  ws = ss_alb_ice_drc     (ns,nr)
-                  gs = asm_prm_ice_drc    (ns,nr)
-               else ! linear interpolation in rsnw
-                  nr = ceiling(rsnw(ksnow)) - 30 + 1            ! hardwired min radius = 30
-                  delr = (rsnw(ksnow) - floor(rsnw(ksnow))) &   ! hardwired delta radius = 1 in table
-                       / (ceiling(rsnw(ksnow)) - floor(rsnw(ksnow))) ! denom always = 1
+               else
+                  ! check whether rsnw_snicar_chk are integers with delta of one, makes search faster
+                  if (rsnw_snicar_chk == 'unknown') then
+                     rsnw_snicar_chk = 'delta1'
```
eclare108213, line 4819: why not initialize rsnw_snicar_chk = 'delta1' above?  Or assume it's something other than delta1 and rearrange the logic?

columnphysics/icepack_shortwave.F90
apcraig: These loops are checking too different conditions.  The first is that the array is basically all integers (to roundoff-ish).  The second is that the delta between two elements in the array is 1.0 (to roundoff-ish).  Both of those need to be true for the "shortcut" to work.

columnphysics/icepack_shortwave.F90
apcraig: You only want to do this check once.  So it's initialized to "unknown".  Then on first entry, it decides if it's "delta1" or "mixed" and then these checks are never done again.  This is like a logical (first_call), but we need more than two states.  We could have a first_call logical and then we could have a string that defaults to "mixed" then resets on first_call to "delta1" if it meets the criteria.  Would that be clearer?

columnphysics/icepack_shortwave.F90
eclare108213: ok thx

columnphysics/icepack_shortwave.F90
eclare108213: that would be clearer.  Even better, could this check be moved into an initialization routine somewhere in columnphysics?

apcraig: OK, let me see what I can do.

apcraig: OK, I refactored this part.  I added a new variable, rsnw_datatype, to icepack_shortwave_data.F90, so that variable is set at initialization.  Then the value of that variable can be used to manage the rsnw interpolation logic.

apcraig:
Latest full test suite results, 

https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#741f301381b0281d2bcbb726968153eb3178c514

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#9c2f26c39fd251a9283c8299c3ca1941e9a1ce7d

columnphysics/icepack_shortwave.F90
eclare108213: This is much clearer now, thank you.

eclare108213 (approved):

dabail10 (commented):
This all makes sense to me and seems fine given that it is bfb.

PR #474, opened 2023-10-27
use -puny in denominator of scaling ratio for flux limiting 

eclare108213:
    Fix the sign of puny in the denominator of the ratio used to scale heat fluxes
- [x] Developer(s): 
@njeffery 
Bug fix tested in E3SM by @njeffery.  
    - [ ] bit for bit
    - [x] different at roundoff level
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
Bottom and lateral heat fluxes are limited by the total amount of heat available for melting, and all of these terms are negative.  In particular, fbot+fside<0. Therefore the value of puny needed to keep the denominator from being too small should also be negative.

### DIFF
--- a/columnphysics/icepack_therm_vertical.F90
+++ b/columnphysics/icepack_therm_vertical.F90
@@ -638,7 +638,7 @@
       ! FYI: fside is not yet correct for fsd, may need to adjust fbot further
       !-----------------------------------------------------------------
 
-         xtmp = frzmlt/(fbot + fside + puny)
+         xtmp = frzmlt/(fbot + fside - puny)
          xtmp = min(xtmp, c1)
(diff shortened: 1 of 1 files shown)


proteanplanet (approved):
I have examined the code and this is indeed a bug that needs to be fixed.

njeffery (approved):
Prevents a possible NaN when abs(fbot+fside) = puny. Tested in a 5 year DTEST with round-off level differences.

apcraig (approved):

apcraig:
Just want to add a note that this changes answers in all cases.

PR #494, opened 2024-07-25
Alternative congelation formulation following Plante et al. 2024

### LINKED ISSUE #481: Salinity and enthalpy for new congelation growth.
This code was noted to be an issue my Mathieu Plante (ECCC). This likely overestimates the salt content and enthalpy (set to the same values as sea water) for the new congelation growth (dhi). This results in too much heat being extracted from the ocean mixed layer and hence more frazil production.

https://github.com/CICE-Consortium/Icepack/blob/f6ff8f7c4d4cb6feabe3651b13204cf43fc948e3/columnphysics/icepack_therm_vertical.F90#L1257-L1265

eclare108213:
Draft for Plante et al. to confirm that these changes are all that are needed, separate from their flooding/snow-ice and forcing modifications.  I still need to add tests to the test suite and update the Icepack documentation, then implement and test in CICE. Suggestions for better congel_freeze option labels are welcome.

I realigned some columns while working on this.  Ignore white space when looking at diffs!

Add alternative congelation formulation following Plante et al., 2024.
- [x] Developer(s): 
@mathieuslplante et al., @eclare108213 
`congel_freeze = ‘two-step’` is BFB in the base_suite -- this is currently the default (as before).
`congel_freeze = 'one-step'` is not BFB. Setting `conserv_check = .true.` does not turn up any issues.
A QC comparison for CICE PR [#965](https://github.com/CICE-Consortium/CICE/pull/965) show the differences between `congel_freeze = 'one-step'` and `'two-step'` are not climate changing in CICE standalone simulations.
    - [x] bit for bit (default)
    - [x] more substantial when `congel_freeze = 'one-step'`
    - [ ] Yes
    - [x] No
    - [x] Yes
    - [ ] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

Namelist flag congel_freeze chooses which formulation to use.  The original is ‘two-step’, since only the mushy boundary moves in the first step and the phase change happens in the next step.  Plante et al. (‘one-step’) moves the boundary and performs the phase change in a single time step.

Closes #481

### DIFF
--- a/columnphysics/icepack_parameters.F90
+++ b/columnphysics/icepack_parameters.F90
@@ -162,6 +162,11 @@
          conserv_check = .false.  ! if true, do conservations checks and abort
 
       character(len=char_len), public :: &
+         congel_freeze  = 'two-step'  ! congelation computation
+                                      ! 'two-step' = original formulation
+                                      ! 'one-step' = Plante et al, The Cryosphere, 2024
+
+      character(len=char_len), public :: &
          tfrz_option  = 'mushy'   ! form of ocean freezing temperature
                                   ! 'minus1p8' = -1.8 C
                                   ! 'constant' = Tocnfrz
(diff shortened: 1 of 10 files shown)


mathieuslplante (commented):
I confirm that the new one-step option corresponds exactly to our changes in Plante et al. (2024).

dabail10 (approved):

apcraig:
Will need some changes in Icepack documentation and maybe add a test case.  Also need to update the icepack interface documentation (run ./icepack.setup --docintfc).  CICE will also eventually need a new namelist added.

apcraig:
See also https://github.com/CICE-Consortium/CICE/pull/965

This PR and the CICE PR are ready to merge assuming we are happy leaving the old 'two-step' as the default.  Please confirm that's how we want to proceed.  We can always change the default later.

dabail10:
I did do a fully-coupled test here with the one-step option and the solutions are quite different. Climate changing? Maybe not, but I think this is the correct way to go. We will definitely be using this in CESM3.

Issue #488, opened 2024-05-22
MOSAiC forcing from Xiaochun Wang


eclare108213:
Xiaochun Wang has offered to provide the MOSAiC forcing that he has used for testing Icepack.

eclare108213:
Pinging @xcwang2019
In this issue we can discuss the steps needed to access the forcing data and update the code and scripts to be able to use it.

xcwang2019:
Currently the MOSAiC forcing my group used is organized in two periods (Nov. 2019 to May 2020, and June -July 2020) since there is a gap in radiation observation.  Please let me know what I should  do.

dabail10:
I would like to loop @davidclemenssewall into this.

giuliacast:
I am trying to run Icepack with MOSAiC forcing as well, I would like to be kept in the loop.

davidclemenssewall:
Hi All,

So sorry that it has taken me this long to get to this (started a new job, field campaign, vacation, etc). Xiaochun and Giulia, I'm thrilled that you're working on using MOSAiC data for Icepack development. There is actually a [cross-cutting working group](https://mosaic-expedition.org/the-science-28/cross-cutting_groups/), the MOSAiC Model Forcing Datasets Working Group (MMFD) within the MOSAiC Consortium that is dedicated to developing merged datasets for model development. You and your team members would be more than welcome to join. We hold a monthly virtual meetings (the next is Oct. 16 at 14:30 UTC) and have both a [public ](https://github.com/SCMStandz/SCMStandz_public) and a private github repo for sharing the code 'recipes' to produce merged datasets. Speaking of which, we published a first version of gap-filled MOSAiC datasets (which include data in the May-June gap from ASFS observations) last spring: [https://arcticdata.io/catalog/view/doi%3A10.18739%2FA2GX44W6J](https://arcticdata.io/catalog/view/doi%3A10.18739%2FA2GX44W6J). The datasets are netCDF in the [MDF format](https://gitlab.com/mdf-makers/mdf-toolkit), which came out of the YOPPSiteMIP project. If you would like to get involved with MMFD please email me at david.clemens-sewall@noaa.gov.

On a related note, I actually have a branch of Icepack that is set up to ingest MDF-formatted forcing data. There are a few things that I had been hoping to add to it still, hence why I hadn't submitted a PR. But it is usable as is and so I will plan to go ahead and submit a PR shortly so others can make use of it.

Cheers,
David

[...]

PR #500, opened 2024-09-24
MDF/MOSAiC forcing options

davidclemenssewall:

## PR checklist
    Enable the ingestion of forcing data from the MOSAiC Expedition and provide more flexibility for ice and ocean initialization.
- [X] Developer(s): 
    David Clemens-Sewall
    Icepack base_suite with baseline comparison to main branch:
215 measured results of 215 total results
215 of 215 tests PASSED
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
        - But we probably should add a standard case for the MOSAiC Expedition, to do so we'll need to add the forcing data to the Icepack standard
    - [X] Yes
        - [ ] Yes
        - [ ] No 
The overall goal of these options added to the Icepack driver are to make it easier to use Icepack as a standalone tool forced with observations. These changes have been implemented specifically for the MOSAiC Expedition. But my hope is that they will be readily extendable to other field data (or reanalysis, etc), so that new datasets can simply be used in Icepack, rather than requiring code changes for every new forcing dataset.

Observational data presents the following complications:
1. Missing values. Real observational datasets almost always have missing values (power loss, radiometer dome iced over, etc...). Furthermore, future versions of the observational datasets may fill some of these missing values (e.g., by reprocessing data that previously failed qc).
2. Variable frequencies. Different data streams may be measured at different sampling frequencies. Data might be measured more often than the model's timestep (thus requiring aggregation) or less often (thus requiring interpolation).
3. Different start and end time points. Different data streams may have different start or end times, and updated versions of datasets may extend the time domain.
4. Merged observational datasets are a work-in-progress. The 3 challenges above all touch on the fact that for sea ice field campaigns, it can take years from the end of the campaign until a truly finalized observational dataset is available to force Icepack with (e.g., MOSAiC is 4 years and counting, SHEBA still does not have such a dataset). We still want to be able to use campaign data to support Icepack development during this time, without having to change the Icepack driver every time the forcing dataset is updated.

Additionally, for comparing with field data it is very useful to be able to easily adjust:
1. The initial state of the sea ice and ocean mixed layer.
2. Fixed values for the oceanic heat flux convergence (```qdp```) and sea surface salinity (```sss```).

To address these challenges the PR provides the following features:
1. Precalculate the forcing values at each timestep on initialization (in ```init_forcing```). The current paradigm for forcing data in Icepack is that when ```init_forcing``` is called, a subroutine for the specific dataset (e.g., ```atm_CFS```) stores the forcing data in the ```*_data``` in essentially the same format that the raw data is present in, without timestamp information. It also hardcodes the maximum size that a forcing dataset can be to 8760. Then, at each timestep the ```get_forcing``` subroutine has a code block for each forcing dataset that contains the timestamp information and interpolates the forcing data to the given timestep. Under this paradigm, I could not figure out a graceful way to handle missing values, nor forcing data with higher sampling frequency than the Icepack timestep, nor changing start times of the forcing dataset. This PR adds a new option, ```precalc_forc```, for use with the MOSAiC forcing. When ```precalc_forc = .true.```, the forcing data is aggregated/interpolated to the Icepack timestep in the call to ```init_forcing``` and stored the the ```*_data``` arrays. Then, the ```get_forcing``` subroutine merely indexes into the ```*_data``` arrays.
2. Read forcing data from MDF-formatted netCDF files (Uttal et al. 2024; https://doi.org/10.5194/gmd-17-5225-2024, https://gitlab.com/mdf-makers/mdf-toolkit). I chose this format because its framework seemed well-aligned, but it is still in development so we can add variables to it that we need for the ice model (e.g., mixed layer depth).
3. To handle variable frequencies and missing data, forcing data are first temporally-averaged for each timestep and then interpolated. For a give data variable (```var_data```), The ```MOSAiC_average``` subroutine takes the average of all forcing datapoints within each ```timestep +- 0.5 dt``` excluding missing values and stores the results in ```var_data```. If there is no valid data within 0.5 dt of a given ```timestep``` (e.g., most timesteps if the model timestep is much smaller than the sampling period) then a missing value is placed in ```var_data(timestep)```. Then, the ```MOSAiC_interpolate``` subroutine linearly interpolates missing values in ```var_data```.
4. The option to explicitly set in the namelist the initial ice and snow thicknesses for the slab and ITD grid cells and the SST for the 3 ice grid cells.
5. The option to explicitly set in the namelist fixed ocean mixed layer conditions ```qdp_fixed```, ```sss_fixed```, and ```hmix_fixed```. These are overwritten by forcing data.

Outstanding issues:
 - ~~Automatically detect cadence of the inputs and adjust accordingly (currently the code requires that the atmospheric forcing is at 1 minute cadence and that the oceanic forcing is at daily cadence)~~
 - ~~It is not obvious to a user that setting the **atmospheric** forcing (```atm_data_type = 'clim'```) can overwrite the **oceanic** heat flux convergence.~~
 - ~~The atm_MOSAiC and ocn_MOSAiC subroutines are nearly identical except for the data variables. It would probably be cleaner to combine them into a single subroutine.~~
 - ~~Perhaps input checking should be added to prevent users from setting ```precalc_forc = .true.``` with non MOSAiC forcings.~~
 - ~~Update documentation to include:~~
   - ~~How driver handles MDF formatted forcing~~
   - ~~MOSAiC forcing~~
   - ~~Initial ice and SST conditions in namelist~~
   - ~~Fixed ocean mixed layer conditions in namelist~~
 - ~~Add warning in log for if MOSAiC time domain includes leg 4-5 transition (observatory relocation)~~
 - Add MOSAiC test case
 - Add link to MOSAiC forcing data to wiki


davidclemenssewall:
@apcraig When I try running the base_suite on Windows Subsystem for Linux (-m conda -e linux), I fail the run stage for the conda_linux_smoke_col_1x1_alt03_debug_run1year and conda_linux_smoke_col_1x1_alt04_debug_run1year tests due to a floating-point exception. I've copied the relevant sections from the runlog below. Strangely, this error does not arise in the GH macos-testing. It also does not occur when I run these tests with the main branch locally on WSL. So it seems to be some combination of the platform and the code changes, although I cannot see how these code changes in the driver cause this error in the atmospheric stability. I was hoping you might have a suggestion on how I can fix this. Thanks!

Tail from the icepack runlog:
```
  Initial forcing data year =         2015
  Final   forcing data year =         2015
 Reading /root/icepack-dirs/input/Icepack_data/forcing/SHEBA/open_clos_lindsay.dat                                      


Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7f05dd5e951f in ???
#1  0x557a0eb656c2 in compute_stability_parameter
        at /root/code/CICE/Icepack/columnphysics/icepack_atmo.F90:992
#2  0x557a0eb67de2 in atmo_boundary_layer
        at /root/code/CICE/Icepack/columnphysics/icepack_atmo.F90:276
#3  0x557a0eb65d98 in __icepack_atmo_MOD_icepack_atm_boundary
        at /root/code/CICE/Icepack/columnphysics/icepack_atmo.F90:954
#4  0x557a0eaf4c59 in __icedrv_step_MOD_ocean_mixed_layer
        at /root/code/CICE/Icepack/configuration/driver/icedrv_step.F90:1251
#5  0x557a0ea69c25 in coupling_prep
        at /root/code/CICE/Icepack/configuration/driver/icedrv_RunMod.F90:303
#6  0x557a0ea6d57b in __icedrv_runmod_MOD_ice_step
        at /root/code/CICE/Icepack/configuration/driver/icedrv_RunMod.F90:221
#7  0x557a0ea6d6dc in __icedrv_runmod_MOD_icedrv_run
        at /root/code/CICE/Icepack/configuration/driver/icedrv_RunMod.F90:57
#8  0x557a0ea6944d in icedrv
        at /root/code/CICE/Icepack/configuration/driver/icedrv_MAIN.F90:51
#9  0x557a0ea69753 in main
        at /root/code/CICE/Icepack/configuration/driver/icedrv_MAIN.F90:28
```

apcraig:
I don't know anything about Windows systems, but I just tested your branch on derecho for the "alt04_debug_run1year".  It passed with intel, fails with gnu with exactly the same traceback as above.  So this is a problem in the code modifications.  If you need me to help you debug, let me know.

apcraig:
Doing just some quick debugging, the value of hol at line 992 of icepack_atm.F90 is a NaN.  Going up one line, tstar and qstar are both NaN in the computation of hol.  There are several forcing options in Icepack, is it possible that your changes work for mosaic and a subset of the forcing, but not all forcing settings?

davidclemenssewall:
Thanks for the help! With that info I was able to track down the source of the 'run' error to be due to the initialization of the ```sstdat``` array in icedrv_flux.F90 from the ```sst``` array. One of the changes that I'm proposing with this PR is to allow the user to explicitly set the initial SST, which I think can be useful when looking initializing before freeze-up or during the melt season when the mixed layer is relatively fresh. But I had failed to propagate that change to the sstdat array.

The latest commit now passes all of the 'run' tests on WSL with the GNU compiler. But it is failing the baseline comparison with the main branch for all of the tests that use the ```alt0[1-4]``` namelist options (see failing tests below). It appears to me the shared factor in the tests that failed is that they contain ```ktherm  = 1 conduct = 'bubbly'```. Whereas the other tests either have ktherm = 2 or, if they use Bitz and Lipscomb thermodynamics they use MU71 conductivity.

Any suggestions would be welcome, I'll keep working on this.

Failing test results:
```
(base) root@dsewall-wl-psl:~/code/CICE/Icepack/testsuite.mosaic_forcing_clean# ./results.csh | grep FAIL
FAIL conda_linux_smoke_col_1x1_alt01_debug_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_alt02_debug_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_alt03_debug_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_alt04_debug_run1year compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_alt01 compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_alt02 compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_alt03 compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_alt04 compare icepack.main different-data
 FAIL - test failed
8 of 215 tests FAILED
```

davidclemenssewall:
The most recent commit fixes this issue and all checks now pass with identical output.

apcraig:
Should we add a couple new tests to the test suite for mosaic?  I assume this would require the mosaic forcing datasets be shared on zenodo?  What's the plan for sharing that forcing dataset as well as any mosaic data "for comparisons" with respect to the Consortium?

@davidclemenssewall, there is a bunch of trailing whitespace in your code changes which we try to avoid.  You can clean it up quickly by doing 

sed -i 's/[ \t]*$//' $file

where $file is the file to be cleaned.  I will also check and fix this before the next release.  thanks,

tony.......

davidclemenssewall:
Fixed the trailing whitespace, thanks for the shell command!

I agree that we should have some tests with this. I think at minimum there should be a run with the 'atmmosaic' and 'ocnmosaic' initialized in November 2019 with namelist options to set the initial snow and ice thicknesses and run through July of 2020. I'm not sure if it makes sense to include some tests of different physics options with this forcing as well? I don't think any of the other forcing options provide daily updates to qdp, sss, and hmix so perhaps having the new congelation growth option would be nice to test.

The forcing datasets are currently archived at ADC: https://arcticdata.io/catalog/view/doi%3A10.18739%2FA2GX44W6J. We expect to be updating the MOSAiC forcing and evaluation (or comparison) datasets for the next couple years at least. So I'm not sure if it makes sense to add it to the existing tarball, or add it as another row in the table on this page: https://github.com/CICE-Consortium/Icepack/wiki/Icepack-Input-Data. It is also a little bit different from the existing forcing datasets in that users would be encouraged to use it for scientific results as well as code testing. Perhaps we should have a meeting to discuss?

apcraig:
Thanks @davidclemenssewall.

If we are going to have some mosaic tests in the test suite, then we'll need to have a reasonably stable forcing dataset that we can install on all the test machines.  We'll want to put it on zenodo and have it be part of the Icepack inputdata.  Happy to have a broader discussion about how to best proceed.

@davidclemenssewall, if you could propose one (or a few) test cases that we could add to the test suite, that would be great.  You can use multiple "options" including 'atmmosaic' and 'ocnmosaic' as needed and turn on the appropriate physics.  You can create some set_nml if needed and add them to the base_suite.ts if you'd like.  For our testing, we only really need enough test cases to cover the code productively.  We do not need to cover all the various physics combinations that a user might find reasonable.  We need to be somewhat clever to cover "everything" while still limiting the number of tests as much as possible.

eclare108213:
> Outstanding issues:
> 
> - Automatically detect cadence of the inputs and adjust accordingly (currently the code requires that the atmospheric forcing is at 1 minute cadence and that the oceanic forcing is at daily cadence)

If this is easy to implement, let's do it.  Otherwise I'd rather merge the PR earlier and improve it later.

> - It is not obvious to a user that setting the atmospheric forcing (atm_data_type = 'clim') can overwrite the oceanic heat flux convergence.

This could be noted in the documentation and also as a warning in the diagnostics log file when namelist values are written out.

> - The atm_MOSAiC and ocn_MOSAiC subroutines are nearly identical except for the data variables. It would probably be cleaner to combine them into a single subroutine.

Your choice.

> - Perhaps input checking should be added to prevent users from setting precalc_forc = .true. with non MOSAiC forcings.

Definitely.

I also recommend adding the forcing to our datasets and creating a few options for testing, including one that represents a 'production' configuration.  I will still recommend that our codes be used in coupled configurations to capture important feedbacks, but we can adjust our stern warnings in the Icepack docs to acknowledge the MOSAiC configuration as a tool to be used thoughtfully.

dabail10:
HI all. Let's discuss this at the meeting tomorrow. I can understand that it would be nice to have this as a part of the Icepack forcing dataset on zenodo. However, the National Science Foundation requires the datasets be put somewhere like the Arctic Data Center where they do reside currently. This would mean duplicating the data on zenodo. Is there a way to have a "link" from zenodo to the ADC?

anton-seaice (commented):
Thanks for doing this David! This will be super useful

As Elizabeth noted, it needs some documentation.

And when adding the link to the forcing data to the wiki, I think the DOI link is best:

e.g. https://doi.org/10.18739/A2GX44W6J

configuration/driver/icedrv_forcing.F90
```diff
@@ -20,62 +21,64 @@ module icedrv_forcing
...
-           open_data, &
-           clos_data
-
       character(char_len), public :: &
          atm_data_format, & ! 'bin'=binary or 'nc'=netcdf
          ocn_data_format, & ! 'bin'=binary or 'nc'=netcdf
          bgc_data_format, & ! 'bin'=binary or 'nc'=netcdf
-         atm_data_type,   & ! 'default', 'clim', 'CFS'
-         ocn_data_type,   & ! 'default', 'SHEBA'
+         atm_data_type,   & ! 'default', 'clim', 'CFS', 'MOSAiC'
```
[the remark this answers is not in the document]
davidclemenssewall: Thanks for the suggestions Anton! @eclare108213 and @apcraig, what would you think about making the ```atm_data_type``` and ```ocn_data_type``` for this 'MDF' instead of 'MOSAiC'? Then I would also propose that we make the default path for these forcing data: ```/icepack-dirs/input/Icepack_data/forcing/MDF```.

configuration/driver/icedrv_forcing.F90
```diff
@@ -972,6 +1036,317 @@ subroutine atm_NICE
...
+
+#else
+         call icedrv_system_abort(string=subname//&
+         ' ERROR: atm_data_format = "nc" requires USE_NETCDF', &
+         file=__FILE__,line=__LINE__)
+#endif
+      else
+         call icedrv_system_abort(string=subname//&
+         ' ERROR: only NetCDF input implemented for atm_MOSAiC', &
+         file=__FILE__,line=__LINE__)
```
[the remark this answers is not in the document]
davidclemenssewall: Done. This will be fixed in the next push.

davidclemenssewall:
> > Outstanding issues:
> > 
> > * Automatically detect cadence of the inputs and adjust accordingly (currently the code requires that the atmospheric forcing is at 1 minute cadence and that the oceanic forcing is at daily cadence)
> 
> If this is easy to implement, let's do it. Otherwise I'd rather merge the PR earlier and improve it later.
> 
> > * It is not obvious to a user that setting the atmospheric forcing (atm_data_type = 'clim') can overwrite the oceanic heat flux convergence.
> 
> This could be noted in the documentation and also as a warning in the diagnostics log file when namelist values are written out.
> 
> > * The atm_MOSAiC and ocn_MOSAiC subroutines are nearly identical except for the data variables. It would probably be cleaner to combine them into a single subroutine.
> 
> Your choice.
> 
> > * Perhaps input checking should be added to prevent users from setting precalc_forc = .true. with non MOSAiC forcings.
> 
> Definitely.
> 
> I also recommend adding the forcing to our datasets and creating a few options for testing, including one that represents a 'production' configuration. I will still recommend that our codes be used in coupled configurations to capture important feedbacks, but we can adjust our stern warnings in the Icepack docs to acknowledge the MOSAiC configuration as a tool to be used thoughtfully.

Latest commit addresses these with the exception of the warnings, documentation updates, and MOSAiC testing. Working on those currently...

configuration/driver/icedrv_forcing.F90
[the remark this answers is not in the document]
eclare108213: MDF works for me.  But how will you distinguish between MOSAiC MDF data and some other kind of MDF data?

configuration/driver/icedrv_flux.F90
```diff
@@ -184,6 +187,12 @@ module icedrv_flux
          fswthru_idr , & ! nir dir shortwave penetrating to ocean (W/m^2)
          fswthru_idf     ! nir dif shortwave penetrating to ocean (W/m^2)
 
+      ! fixed ocean mixed layer properties (are overwritten by forcing data)
+      real (kind=dbl_kind), public :: &
+         sss_fixed   , & ! Sea surface salinity (PSU)
+         qdp_fixed   , & ! Deep ocean heat flux (negative upward, W/m^2)
+         hmix_fixed      ! Mixed layer depth (m)
+
```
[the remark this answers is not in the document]
eclare108213: I don't think it's worth the extra work at this point, to make a separate PR.

configuration/driver/icedrv_forcing.F90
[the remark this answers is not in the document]
davidclemenssewall: I believe the only place where it matters to the code that this is MOSAiC MDF data is when we check if the simulation time includes the MOSAiC leg 4 to leg 5 ship repositioning. For that check, we can check whether the forcing filename contains the string 'MOSAiC'.

apcraig:
Looks like this is going forward nicely.  I am mostly comfortable with the proposals thus far.  But I think it's a mistake to have a feature that works if "MOSAIC" is in the filename.  Please don't implement something like that.  We should create a new namelist option if we need to.  For now, we can use MDF.  If something comes along later that's MDF but not Mosaic, we'll create a new option for it called "BLAHBLAH" or "BBMDF" or something.  Or if what we have now is specific to mosaic even though it's MDF, lets call it MOSAIC or MosaicMDF or something.  If the implementation is general for MDF, lets use MDF.  How general or specific is the current implementation, that should drive the name we pick for the namelist setting.

configuration/driver/icedrv_forcing.F90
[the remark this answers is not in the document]
anton-seaice: Thanks!

configuration/driver/icedrv_forcing.F90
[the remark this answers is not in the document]
anton-seaice: Gah - thats an annoying quirk. I don't have a straightforward answer to that. This sounds like a different measurement site though? So should the experiment start again at that point with different inititial conditions ?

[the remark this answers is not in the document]
apcraig: Should we have two different mosaic options.  MOSAIC123 and MOSAIC45 for instance or some other name.  These would be two separate forcing/ic options?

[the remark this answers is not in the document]
davidclemenssewall: I think that's a good idea. I have not worked much with the leg 5 (August-September near the North Pole) data but I will inquire about how we could have an appropriate initial condition.

Based on this discussion, I think it would be cleanest if the data archive contained separate files for legs 1-4 (October-July along transpolar drift) and leg 5 (August-September near the North Pole). I will inquire within the MOSAiC Model Forcing Dataset working group to see whether others agree and if we can update the data archive.

davidclemenssewall:
> Looks like this is going forward nicely. I am mostly comfortable with the proposals thus far. But I think it's a mistake to have a feature that works if "MOSAIC" is in the filename. Please don't implement something like that. We should create a new namelist option if we need to. For now, we can use MDF. If something comes along later that's MDF but not Mosaic, we'll create a new option for it called "BLAHBLAH" or "BBMDF" or something. Or if what we have now is specific to mosaic even though it's MDF, lets call it MOSAIC or MosaicMDF or something. If the implementation is general for MDF, lets use MDF. How general or specific is the current implementation, that should drive the name we pick for the namelist setting.

I think that the right approach here is to break the MOSAiC forcing data into separate MDF files for legs 1-4 and leg 5. Then the forcing code can remain totally general for MDF, and we have two different set_env options for initializing the different legs as you suggested.

xcwang2019:
@davidclemenssewall  @eclare108213 @dabail10  Sorry to reply the PR so late.  Our group has constructed  atmopheric and oceanic forcing files for Icepack using MOSAiC  observations. The hourly forcing file is separated into two periods. The format is txt.   Should I attach these  files here or should I put it on zenodo and provide a link here?  These forcing files were used on Icepack 1.1.0 and there is no code change involved.  Thanks.

eclare108213:
@xcwang2019 Let's move discussion of your forcing data to issue https://github.com/CICE-Consortium/Icepack/issues/488.
@davidclemenssewall is this PR ready for review or are you still working on it?

davidclemenssewall:
@eclare108213 apologies for the delay here. I am waiting on some additional processed precipitation data that I would like to include in the next version of the MOSAiC forcing data. Once I have the updated dataset released I will request review here.

davidclemenssewall:
Hi all,

This is almost ready for review but I'm running into two issues on testing:

1) When I attempt to run the smoke test using the MOSAiC forcing with the debug flag set, I receive a floating point exception when I try to open the netcdf file (`nf90_open(...`). I've copied the error below It appears that this might be an issue with the underlying netcdf fortran library (I found another thread on this here: https://support.unidata.ucar.edu/archives/netcdf/msg15037.html). When I run without the debug flag the test passes fine. I've contacted the person who asked that support question to see if there was a resolution, but I was curious if anyone here has run into a similar issue?

2) The standard restart test doesn't work for the MOSAiC forcing because drift 1 of the MOSAiC campaign only ran from October 2019 to July 2020. Would it be reasonable to change the restart test such that the restart point is at some point within `npt` (e.g., the restart point could be halfway through the length of the full run)? Or perhaps we should have a third test option (like 'restart_half') that would do this and leave the current restart test alone.

The ReadTheDocs build also failed on this push, but I think that was just a hiccup from the RTD server. The documentation builds fine when I restarted it. See here: https://davidclemenssewall-icepack.readthedocs.io/en/mosaic_forcing_clean/.

I have uploaded version 2 of the MOSAiC Model Forcing Datasets, including the changes requested by the consortium, to the Arctic Data Center. In case you want to test them before they are published, I've put them in a google drive folder here: https://drive.google.com/drive/folders/1xGAeCJDwbvkXJkA-3DZXyr75GOGnoZa4?usp=sharing. Just let me know if you need access.

Cheers,
David

Error when attempting to open netcdf file in debug mode:
```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7fc0a1b4951f in ???
        at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#1  0x7fc0a18b1b72 in ???
#2  0x7fc0a1840cba in ???
#3  0x7fc0a18d01e2 in ???
#4  0x7fc0a1627a22 in ???
#5  0x7fc0a16c4385 in ???
#6  0x7fc0a21a0b9f in ???
#7  0x7fc0a21a952d in ???
#8  0x7fc0a2124213 in ???
#9  0x7fc0a2126359 in ???
#10  0x7fc0a21264d4 in ???
#11  0x7fc0a202ea7b in ???
#12  0x7fc0a20d5fd7 in ???
#13  0x559b0715f35b in atm_mdf
        at /root/code/CICE/icepack/configuration/driver/icedrv_forcing.F90:1097
#14  0x559b07174ed6 in __icedrv_forcing_MOD_init_forcing
        at /root/code/CICE/icepack/configuration/driver/icedrv_forcing.F90:195
#15  0x559b071347c2 in __icedrv_initmod_MOD_icedrv_initialize
        at /root/code/CICE/icepack/configuration/driver/icedrv_InitMod.F90:137
#16  0x559b071348dd in icedrv
        at /root/code/CICE/icepack/configuration/driver/icedrv_MAIN.F90:45
#17  0x559b07134be8 in main
        at /root/code/CICE/icepack/configuration/driver/icedrv_MAIN.F90:28
```

anton-seaice:
> 1. When I attempt to run the smoke test using the MOSAiC forcing with the debug flag set, I receive a floating point exception when I try to open the netcdf file (`nf90_open(...`). I've copied the error below It appears that this might be an issue with the underlying netcdf fortran library (I found another thread on this here: https://support.unidata.ucar.edu/archives/netcdf/msg15037.html). When I run without the debug flag the test passes fine. I've contacted the person who asked that support question to see if there was a resolution, but I was curious if anyone here has run into a similar issue?

If this is using a netcdf4 file, there was an issue with the underlying `hdf5` library in some versions last year that caused this sort of error. I think it only occurred in `hdf5` versions `1.14.3` and was also patched in the spack build. See https://github.com/HDFGroup/hdf5/issues/3831

eclare108213:
@davidclemenssewall, is this PR ready for review now?

apcraig:
I will start some testing.  There are not changes in columnphysics.  But a few questions

- Do we want to make changes in CICE to support similar namelist/initialization options (*_init*, *_fixed and precalc namelist)?
- Where are the mosaic forcing dataset?  Looks like they have NOT been copied onto NCAR's CICE-Consortium inputdata space?  Have they been placed on zenodo?  I am downloading the datasets off the google drive now from @davidclemenssewall.  Are we able to use and share those datasets now?
- What is the appropriate mosaic test setup?  There are two new namelist options, atmmosaic and ocnmosaic.  I assume both need to be turned on together for the mosaic test?  Or can we test each option separate?  Are there any other options that need to be turned on with the mosaic test?

apcraig:
Testing looks good.  I have added the mosaic forcing datasets from the google drive to the NCAR/derecho inputdata area and have run full Icepack test suites on derecho with success.  I added a mosaic test to my sandbox that sets atmmosaic,ocnmosaic,ionetcdf.  That builds and runs with all the derecho compilers.  I would like to add that test to the PR.  Can we do that thru @davidclemenssewall's branch or should I pull the branch onto my fork and replace this PR with a PR from my branch?

apcraig:
Just pinging again to see how @davidclemenssewall, @dabail10, @eclare108213 want to proceed.  Can we put the MOSAIC forcing on zenodo?  What's the best process for adding a mosaic test, I have one ready?  

Also, was just looking thru the comments on this PR.  There was some discussion about having two mosaic cases, one for legs 1-4 and another for leg 5.  I only see one mosaic setup at present time.  What's the status of the current forcing and mosaic setup.  Is it just legs 1-4?

NickSzapiro-NOAA:
If it helps, MOSAiC forcing is already published to Arctic Data Center (https://doi.org/10.18739/A2WD3Q35Z) into drift1 and drift2 legs

apcraig:
Thanks @NickSzapiro-NOAA, that is very helpful.  So, the dataset is in the public domain and already has a doi.  It looks like we are using the same six MOSAIC_ datasets.  There are drift1 and drift2 datasets, those seem to be legs 1-4 and leg 5.  Our current test case uses the drift1 dataset.  I guess we could setup another test for drift2, just making sure we set appropriate initial conditions, start date, and run length.  It would probably be good to do that.  Who can help with that?

So, it'd be nice not to republish the Mosaic forcing dataset on zenodo for the Consortium.  Is there a way to point the Consortium community to the Arctic Data Center dataset in a simple way?

NickSzapiro-NOAA:
> So, it'd be nice not to republish the Mosaic forcing dataset on zenodo for the Consortium. Is there a way to point the Consortium community to the Arctic Data Center dataset in a simple way?

Not so clean and maybe someone knows better, but this gets all files:
` wget -O mosaic.zip https://arcticdata.io/metacat/d1/mn/v2/packages/application%2Fbagit-1.0/resource_map_doi%
3A10.18739%2FA2WD3Q35Z`

or individual netcdf files using their download links, like:
`curl --output MOSAiC_atm_drift2_precip_MDF_20200827_20200918.nc https://arcticdata.io/metacat/d1/mn/v2/object/urn%3Auuid%3A332e7185-9b5b-4ffe-98a6-e99186be2276`

eclare108213:
We have wiki pages that point to our existing data sets on zenodo.  We can just put the link to the MOSAiC data doi there.  Then people can download them however they want.  Do we need all of the data or just some of it?

apcraig:
Great, lets use the existing Arctic Data website and existing doi to reference the Mosaic dataset.  I think we may only need a few of the files, but the entire dataset is less than 70 Mb, which is large compared to the other Icepack datasets, but is still pretty small.  Another question as I look over things, it seems the new atm precip datasets are not being used by Icepack, only the atm stakes and ocn MDF files are used.  I wonder if this is something that still needs to be addressed?

NickSzapiro-NOAA:
@davidclemenssewall really is the expert here. 

There is a working group for these data (https://github.com/SCMStandz/SCMStandz_public) ... can make issue there or raise at next month's meeting. Also, I imagine some folks are at ASSW 2025 if you are nearby.

davidclemenssewall:
Hi all, sorry for my slowness here.

I have updated the atmmosaic option to use the precip forcing file instead of the stakes files and added a MOSAiC drift 1 test to the base_suite. I also confirmed that the base_suite passes all tests on conda linux.

The precip file has our best estimate of the actual precipitation input, whereas the pseudo-precipitation used in the stakes files has been tuned to reproduce the observed snow depth evolution at the manual mass balance ('stakes') sites. So the stakes pseudo-precipitation is combining the effects of true precip and snow redistribution. In the long term, I think it is better for people to use a true precipitation dataset and tune the snow redistribution parameters to appropriately reproduce the snow depths at the mass balance sites. In the current test case, snow redistribution is not feasible because we do not have an appropriate dynamics forcing yet.

I am a bit loath to add a drift 2 test case without also adding an ice initial conditions file because when drift 2 measurements started in August, there were well-developed melt ponds. For now I propose we stick with just the drift 1 test case. In the future we can add a freeze-up focused test case.

If everyone agrees with that, then the only files that need to be downloaded for the Icepack test case are `MOSAiC_atm_drift1_precip_MDF_20191015_20200731.nc` and `MOSAiC_ocn_drift1_MDF_20191006_20200731.nc`.

I think that covers everything and this is ready to merge, but please let me know if I missed anything.

eclare108213:
@davidclemenssewall Thank you for adding the new test case and for the explanation of the different precip data sets. This is very helpful. 

I suggest adding comments to the documentation and wiki page along the lines of @davidclemenssewall's note above, to help clarify what the differences are, in case someone is interested in running with effective wind redistribution or comparing the two.  This is a suggestion -- reword or reformat as needed:

In the doc (`ug_running.rst`):
> The precip file `MOSAiC_atm_drift1_precip_MDF_20191015_20200731.nc` has our best estimate of the actual precipitation input, whereas the pseudo-precipitation used in the stakes file `MOSAiC_atm_drift1_stakes_snow_fyi_MDF_20191015_20200731.nc` has been tuned to reproduce the observed snow depth evolution at the manual mass balance ('stakes') sites. Therefore the stakes pseudo-precipitation effectively combines the effects of true precipitation and snow redistribution. 

In the [data wiki page](https://github.com/CICE-Consortium/Icepack/wiki/Icepack-Input-Data) (@apcraig or I will make this change when the PR is merged, and include a direct link):

>    MOSAiC forcing data is available directly from the Arctic Data Center.  Initial conditions for drift 2 (leg 5) data have not yet been implemented in Icepack.

>   Standard MOSAiC forcing data:
>   - MOSAiC_atm_drift1_precip_MDF_20191015_20200731.nc  -  observed atmospheric data for drift 1 (legs 1-4)
>   - MOSAiC_ocn_drift1_MDF_20191006_20200731.nc. - observed ocean data for drift 1 (legs 1-4)
>
>   MOSAiC forcing data with pseudo-precipitation (optional):
>   - MOSAiC_atm_drift1_stakes_snow_fyi_MDF_20191015_20200731.nc - observed atmospheric data for drift 1, with snowfall data tuned to match observations at the manual mass balance 'stakes' sites

apcraig:
Thanks @davidclemenssewall, that's great.  I will test the latest changes on Derecho, maybe you can update the documentation as suggested, then we can merge.

davidclemenssewall:
Thanks @eclare108213 and @apcraig. I have updated the documentation to provide more details about the precipitation options and the initial conditions for drift 1.

apcraig (commented):
The MDF option needs to be added to the documentation under atm_data_type and ocn_data_type in the namelist description.  Please check that all the new namelists are documented there correctly.

apcraig:
Test results look good, https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#c4dc7f2a69333e247985d1168cee9d4d4615fe2e.

davidclemenssewall:
> The MDF option needs to be added to the documentation under atm_data_type and ocn_data_type in the namelist description. Please check that all the new namelists are documented there correctly.

Good catch! I have now updated this, and checked that the other new namelist options are documented in those tables.

apcraig (approved):

apcraig:
Thanks @davidclemenssewall, everything looks good and I have approved the PR.  I will merge tomorrow (April 1) unless I hear otherwise.

configuration/driver/icedrv_forcing.F90
```diff
@@ -161,6 +192,7 @@ subroutine init_forcing
       if (trim(atm_data_type(1:4)) == 'clim')  call atm_climatological
       if (trim(atm_data_type(1:5)) == 'ISPOL') call atm_ISPOL
       if (trim(atm_data_type(1:4)) == 'NICE')  call atm_NICE
+      if (trim(atm_data_type(1:6)) == 'MDF')   call atm_MDF
```
eclare108213, line 195: Shouldn't this be atm_data_type(1:3) ?

configuration/driver/icedrv_forcing.F90
```diff
@@ -176,6 +208,7 @@ subroutine init_forcing
 
       if (trim(ocn_data_type(1:5)) == 'ISPOL') call ocn_ISPOL
       if (trim(ocn_data_type(1:4)) == 'NICE')  call ocn_NICE
+      if (trim(ocn_data_type(1:6)) == 'MDF') call ocn_MDF
```
eclare108213, line 211: Shouldn't this be ocn_data_type(1:3) ?

eclare108213:
Let me finish reviewing the code before merging, thanks.

configuration/scripts/options/set_nml.atmmosaic
```diff
@@ -0,0 +1,17 @@
...
+npt             = 6132
+precalc_forc    = .true.
+hi_init_slab    = 0.55
+hsno_init_slab  = 0.07
+hbar_init_itd   = 1.5
+hsno_init_itd   = 0.15
+qdp_fixed       = -1.0
+sst_init        = -1.8
+sss_fixed       = 32.0
+hmix_fixed      = 45.0
```
eclare108213, line 17: It looks like there isn't a line return at the end of the new set_nml files.  That has sometimes been a problem on some architectures or compilers.

doc/source/developer_guide/dg_driver.rst
```diff
@@ -40,7 +40,7 @@ Overview
 
 The icepack driver exists to test the column physics.  At the present time, it is hardwired
 to run 4 different gridcells on one processor with the same forcing used for all gridcells.  
-There is no MPI and no threading built into the icepack driver.  There is limited IO capabilities,
-no history files, and no netcdf restart files.  The model generally runs very quickly.
+There is no MPI and no threading built into the icepack driver.  There is limited IO capabilities.
```
eclare108213, line 43: As long as you're fixing this line, please also fix the pre-existing grammar error:  "There are ... capabilities."  Thanks!

doc/source/user_guide/ug_implementation.rst
```diff
@@ -118,17 +118,74 @@ will be described in more detail in the :ref:`tabnamelist`.
...
+at each timestep the ``get_forcing`` subroutine has a code block for 
+each forcing dataset that contains the forcing's time basis and 
+interpolates the forcing data to the given timestep. The forcing data
+that is available in 'bin' format are intended merely for code testing,
+not scientific results.
+
+If ``precalc_forc`` = .true., ``atm/ocn_data_type`` = 'MDF', and 
+``atm/ocn_data_format`` = 'nc', then ``init_forcing`` reads data from
+netCDF files formatted according to the Merged Data File (MDF),
+conventions which includes timestamp information. During initialization,
```
eclare108213, line 161: The comma should be after 'conventions' instead of (MDF)

doc/source/user_guide/ug_running.rst
```diff
@@ -736,6 +736,46 @@ cases.  Current filenames can be found in the options scripts in
...
+    pseudo-precipitation effectively combines the effects of true precipitation and snow redistribution.
+    It is recommended to use the precip file for general testing. However, to specifically reproduce the
+    observed snow depth at the manual mass balance sites with the precip forcing, one would need also need to 
+    accurately reproduce the wind-driven snow redistribution. See :cite:`Raphael24` for further details on the
+    manual mass balance measurements.
+    
+    Oceanic forcing fields consist of mixed layer salinity (PSU),
+    mixed layer depth (m), turbulent heat fluxes over the halocline and thermocline (:math:`W/m^2`),
+    mixed layer temperature (K), ocean-ice friction velocity (m/s), and sea ice drift speed (m/s). Note,
+    icepack's driver currently lacks a mechanism to use the ocean-ice friction velocity and/or sea ice
```
eclare108213, line 770: I prefer to capitalize when using it as a proper noun (the name of the model).  We use the lower-case version as the submodule link.

eclare108213: This probably isn't consistent elsewhere in the code...  no need to fix anywhere else, just trying to move in that direction.

doc/source/user_guide/ug_running.rst
```diff
@@ -736,6 +736,46 @@ cases.  Current filenames can be found in the options scripts in
...
+    drift speed in the thermodynamics calculations. Additionally, caution should be used when using the
+    ocean mixed layer temperatures directly. For much of the experiment, Icepack's parameterization
```
eclare108213, line 772: For clarity, "... caution should be used when using the modeled ocean mixed layer temperatures."  (The word "directly" threw me.)

doc/source/user_guide/ug_running.rst
```diff
@@ -736,6 +736,46 @@ cases.  Current filenames can be found in the options scripts in
...
+    of the ocean mixed layer freezing point differs from the true freezing point by ~0.5 C.
```
eclare108213, line 773: Nothing to change here, but do you have any idea why it differs by so much when you're using MOSAiC forcing data both above and below?

eclare108213 (commented):
I have several minor requests for changes, otherwise this is great.  Thank you.

One general concern:  are the new namelist parameters coded in such a way that they will default to the original values if a driver code (like CICE) does not implement them in its namelist file?  It's probably not an issue since those codes probably don't use Icepack's slab / ITD / etc initializations. What about for Icepack itself, if someone updates their code but continues using an older icepack_in?

apcraig:
@eclare108213, will do.  Will wait for your review.  Thanks.

davidclemenssewall:
This all sounds good, I'm away from my computer the rest of today but I
will fix these things tomorrow evening.

apcraig:
One other thought.   There is an atm mosaic and an ocn mosaic namelist setting.  Do we ever expect to use one without the other?  If not, I suggest we combine them into a single mosaic namelist setting, get rid of the two, and update the test.

eclare108213:
Sincethe two are already implemented, my inclination is to leave them both in.  This might allow mix-and-match testing among data sets (?).

configuration/driver/icedrv_forcing.F90
davidclemenssewall: Good catch, fixed.

configuration/driver/icedrv_forcing.F90
davidclemenssewall: Fixed.

configuration/scripts/options/set_nml.atmmosaic
davidclemenssewall: Fixed.

doc/source/developer_guide/dg_driver.rst
davidclemenssewall: Done.

doc/source/user_guide/ug_implementation.rst
davidclemenssewall: Fixed.

doc/source/user_guide/ug_running.rst
davidclemenssewall: Done. I'll try to remember to update this if/when I see it elsewhere. We're trying to have a few new people run Icepack as part of a sea ice field school next week and so I may have more suggested updates to the documentation based on their feedback.

doc/source/user_guide/ug_running.rst
davidclemenssewall: Done.

doc/source/user_guide/ug_running.rst
davidclemenssewall: This is something I have a note to myself to look into. The short version is that the mushy physics parameterization of the salinity-dependent freezing point differs from the true salinity-dependent freezing point (or at least the TEOS-10 equations, which are consistent with the MOSAiC ocean measurements). Why that is, and how much it impacts results, are both things I would like to investigate further at some point.

davidclemenssewall:
> I have several minor requests for changes, otherwise this is great. Thank you.
> 
> One general concern: are the new namelist parameters coded in such a way that they will default to the original values if a driver code (like CICE) does not implement them in its namelist file? It's probably not an issue since those codes probably don't use Icepack's slab / ITD / etc initializations. What about for Icepack itself, if someone updates their code but continues using an older icepack_in?

In `icedrv_init.F90`, subroutine `input_data`  the new namelist parameters all are given default values that are the same as what they were in the original code. Icepack does run successfully using an old icepack_in file from before the namelist changes were made.

Regarding the two separate namelist settings. During testing it was relatively common for me to use just the atmmosaic setting and then use the namelist ocean options to explore the sensitivity to varying the oceanic heat flux convergence and salinity. I don't know that that is a common usage, but it is why I set it up that way. I don't think I have ever used the ocnmosaic setting without the atmmosaic setting.

doc/source/user_guide/ug_running.rst
eclare108213: Super, thank you.

doc/source/user_guide/ug_running.rst
eclare108213: Flagging @proteanplanet on this comment, since he also mentioned TEOS-10 yesterday in relation to sea ice.

eclare108213 (approved):
Fantastic, thanks!

eclare108213:
Just so we don't forget about this part:

> In the [data wiki page](https://github.com/CICE-Consortium/Icepack/wiki/Icepack-Input-Data) (@apcraig or I will make this change when the PR is merged, and include a direct link):
> 
> > MOSAiC forcing data is available directly from the Arctic Data Center.  Initial conditions for drift 2 (leg 5) data have not yet been implemented in Icepack.
> 
> > Standard MOSAiC forcing data:
> > 
> > * MOSAiC_atm_drift1_precip_MDF_20191015_20200731.nc  -  observed atmospheric data for drift 1 (legs 1-4)
> > * MOSAiC_ocn_drift1_MDF_20191006_20200731.nc. - observed ocean data for drift 1 (legs 1-4)
> > 
> > MOSAiC forcing data with pseudo-precipitation (optional):
> > 
> > * MOSAiC_atm_drift1_stakes_snow_fyi_MDF_20191015_20200731.nc - observed atmospheric data for drift 1, with snowfall data tuned to match observations at the manual mass balance 'stakes' sites

xcwang2019:

PR #503, opened 2024-10-02
Add dorebin and docleanup optional arguments to icepack_step_ridge

### LINKED ISSUE #478: ridge_ice no longer public
Hi all, some of us in the SIS2 community have been calling Icepack's ridge_ice from SIS2. I saw that it became private as of Icepack 1.3.4 and started working on updating how we call Icepack's ridging routines. The public icepack_step_ridge calls both ridge_ice and cleanup_itd. I'm not opposed to the idea of cleaning up the itd in a mass and energy conserving fashion, but something you are doing doesn't play well with SIS2.

* It flat out doesn't work when we use a process mask_table
* Without a mask_table it makes the coupled ice-ocean model unbelievably slow, slower than the simple lack of the mask_table can explain.

Can I ask that ridge_ice be made public again?

apcraig:
## PR checklist
    Add dorebin and docleanup optional arguments to icepack_step_ridge
- [X] Developer(s): 
    kshedstrom, apcraig
    Tested Icepack and CICE on Derecho without the optional arguments, with the optional arguments set to true and with the optional arguments set to false.  Icepack remained bit-for-bit in all cases (probably too few gridcells), CICE changed answers when the optional arguments were set to false.  Otherwise results are bit-for-bit, but will do more testing before merging.  Ran full test suites on Derecho with intel and gnu for Icepack and CICE just before merging, results look fine.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Add dorebin and docleanup optional arguments to icepack_step_ridge to support NOT calling cleanup_itd and NOT calling rebin in the ridge_ice.
Closes #478

Rename limit_aice_in argument in ridge_ice to limit_aice.  This argument is not used in Icepack or CICE at the moment.

Based on https://github.com/ESMG/Icepack/tree/optional_cleanup

Update interface document

### DIFF
--- a/columnphysics/icepack_mechred.F90
+++ b/columnphysics/icepack_mechred.F90
@@ -1740,7 +1740,9 @@
                                     araftn,       vraftn,        &
                                     aice,         fsalt,         &
                                     first_ice,    fzsal,         &
-                                    flux_bio,     closing, Tf )
+                                    flux_bio,     closing,       &
+                                    Tf,                          &
+                                    docleanup,    dorebin)
 
       real (kind=dbl_kind), intent(in) :: &
          dt           ! time step
@@ -1815,12 +1817,20 @@
       logical (kind=log_kind), dimension(:), intent(inout) :: &
          first_ice    ! true until ice forms
 
+     logical (kind=log_kind), intent(in), optional ::   &
+         docleanup, & ! if false, do not call cleanup_itd (default true)
+         dorebin      ! if false, do not call rebin in cleanup_itd (default true)
+
 !autodocument_end
 
       ! local variables
 
       real (kind=dbl_kind) :: &
          dtt          ! thermo time step
+
+      logical (kind=log_kind) ::   &
+         ldocleanup, &! if true, call cleanup_itd
+         ldorebin     ! if true, call rebin in cleanup_itd
 
       logical (kind=log_kind), save :: &
          first_call = .true.   ! first call flag
@@ -1841,6 +1851,17 @@
          endif
       endif
 
+      if (present(docleanup)) then
+         ldocleanup = docleanup
+      else
+         ldocleanup = .true.
+      endif
+
+      if (present(dorebin)) then
+         ldorebin = dorebin
+      else
+         ldorebin = .true.
+      endif
 
       !-----------------------------------------------------------------
       ! Identify ice-ocean cells.
@@ -1880,8 +1901,9 @@
       !  categories with very small areas.
       !-----------------------------------------------------------------
 
-      dtt = dt * ndtd  ! for proper averaging over thermo timestep
-      call cleanup_itd (dtt,                  hin_max,          &
+      if (ldocleanup) then
+         dtt = dt * ndtd  ! for proper averaging over thermo timestep
+         call cleanup_itd(dtt,                hin_max,          &
                         aicen,                trcrn,            &
                         vicen,                vsnon,            &
                         aice0,                aice,             &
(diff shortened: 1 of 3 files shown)


eclare108213:
> Do we need new namelist (in Icepack and CICE) to allow users to turn the features on and off or will this just be hardcoded in the calls when needed?

Since we generally would not recommend these flags be set to false, I do not think they should be exposed in namelist.

eclare108213:
Since it's pertinent, please also remove the reference to a boundary call mentioned [here](https://github.com/CICE-Consortium/Icepack/issues/478#issuecomment-1822093194).  It's around line 742 in ice_itd.F90.  I looked in the old svn repository -- that comment is ANCIENT (more than 20 years).

kshedstrom:
I don't understand it, but I can reproduce my old answers if I add yet a third optional argument to icepack_step_ridge. If the third one is false, I call ridge_ice without the opening and closing arguments, forcing it to call ridge_prep. I thought I was doing all the same computations on the SIS2 side, but something is still different. If chinook ever comes back, I'll have a better shot at debugging it.

kshedstrom:
I figured it out - it's all in the order of operations. I approve this PR.

apcraig:
@kshedstrom Great to hear you figured out what's going on.  So do you need both of these optional arguments to make things work?  Do you understand why?

kshedstrom:
No, I only need the one, but I can get three different answers by having both. I haven't talked to the GFDL folks about this and maybe down the road we'll be able to pick which answer we want, but so far we've only tested it in long runs without any itd_cleanup at all.

columnphysics/icepack_mechred.F90
```diff
@@ -1841,6 +1851,17 @@ subroutine icepack_step_ridge(dt,           ndtd,          &
...
+      if (present(docleanup)) then
+         ldocleanup = docleanup
+      else
+         ldocleanup = .true.
+      endif
+
+      if (present(dorebin)) then
+         ldorebin = dorebin
+      else
+         ldorebin = .true.
```
eclare108213, line 1863: `ldorebin` does not appear to be used in this subroutine

kshedstrom:
It is passed through to cleanup_itd.

eclare108213 (approved):
Ah, now I see `ldorebin` in the `cleanup_itd` argument list.  Thanks for the clarification. I'm still a bit hesitant to approve these changes without understanding what it is about the driver that requires them.  I'm worried about nonphysical results that might not be noticed or accounted for.  But I will approve, since these changes seem important for the GFDL model, and because these options are buried in the code where others are less likely to thoughtlessly experiment with them. If they turn out to be helpful for other groups, then we can expose them in namelist.  I highly recommend the GFDL model's developers look closely at how their model handles ice thicknesses that lie outside of their thickness bins, especially considering @kshedstrom's comment that 
> I only need the one, but I can get three different answers by having both.

apcraig:
@kshedstrom @eclare108213 I am a little concerned that we don't fully understand why we need these changes and whether the science results are valid.  I'm willing to merge this mostly because the feature is pretty well hidden, although the new arguments will appear in our Icepack interface documentation.  @kshedstrom please confirm again that you want these changes as currently implemented in the PR to be merged and that you will continue to validate and report back any results to us.  Thanks.

kshedstrom:
Yes, this is indeed what I want, at least for now. In the MOM6 world having the ability to reproduce old answers can be important and we need one flag for that. I will let you know about the rest as I learn more.

apcraig:
I ran full test suites on Derecho with intel and gnu for Icepack and CICE, tests look good.

PR #504, opened 2024-11-09
hocn calculation and flushing velocity issue

davidclemenssewall:

## PR checklist
    Fixes two small bugs related to how the mass of melt pond water impacts melt pond drainage and synchronizes 'apnd' nomenclature.
- [X] Developer(s): 
    @davidclemenssewall 
    @dabail10 @eclare108213 @apcraig 
Ran the Icepack base_suite on conda_linux with a baseline comparison to a baseline from the latest 'main' branch. All tests passed except the comparison tests for cases that used mushy thermodynamics and level or topo ponds (see below). These tests were expected to fail due to the nature of this code change. I have not been able to test the impacts of these changes in CICE because the current icepack main branch fails to compile in CICE.
FAIL conda_linux_smoke_col_1x1_diag1_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_bgcispol_debug compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_bgcnice_debug compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_zaero compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year_swredist compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_dt30min_leap_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_dyn_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_fsd12_run1year_short compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_fsd1_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year_snw30percent_snwgrain compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year_snwitdrdg compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_calcdragio_debug_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_modal_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_fluxopenw_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_dyn_fluxopenw_run1year compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year_snicartest compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_debug_run1year_snicar compare icepack.main different-data
FAIL conda_linux_smoke_col_1x1_congel_debug_run1year compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_debug compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_diag1 compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_pondlvl compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_pondtopo compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_bgcispol compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_bgcnice compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_snwgrain_snwitdrdg_zaero compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_isotope compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_saltflux compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_dyn compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_fsd12_short compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_snwgrain_snwitdrdg compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_modal compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_fluxopenw compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_snicartest compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_snicar compare icepack.main different-data
FAIL conda_linux_restart_col_1x1_congel compare icepack.main different-data
36 of 215 tests FAILED
    - [ ] bit for bit
    - [X] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
Currently, there are two bugs in how icepack computes the height of the ocean above the ice base in icepack_therm_mushy.F90 line 3141:

```hocn = (ice_mass + hpond * apond * rhow + hsn * rhos) / rhow```

First, ```rhow``` is the density of ocean water, but ponds that depress the ice surface are generally found to be fresh water. Thus, ```rhow``` in the numerator should be replaced by ```rhofresh```.
Second, ```apond``` here is the category pond area tracer, however it is being treated as the category pond area fraction (see footnote below). For the level-ice scheme, the category pond area fraction is ```apnd*alvl```.
This PR fixes both issues. There were also offsetting errors in the ```flushing_velocity``` and ```flush_pond``` subroutines that have been fixed.

I believe that the second bug was made more likely by using inconsistent naming conventions for the pond area tracer and pond area fraction. In icepack_meltpond_lvl.F90 and other parts of the code base, it appears that the general convention is to use ```apnd``` to denote the pond area tracer and ```apondn``` for the category or full-grid-cell pond area fraction (i.e., ```apondn = apnd*alvl``` for the level pond scheme). I propose that we follow this convention throughout the code base, such that ```apnd``` always refers to the pond area tracer. This PR makes these changes in the mushy and vertical thermodynamics modules.

Footnote: there is an alternate interpretation of using just the pond area tracer in the numerator for the level pond scheme which is that the level areas of an ice category separated from the deformed areas by borders of zero flexural rigidity (i.e., the thicker, pond-free, deformed ice cannot provide any upward buoyancy force to the level, ponded areas). If this is the intention, then I think we need to also use the level ice volume (```vlvl```) in the calculation of ice mass and the level-ice snow depth for consistency.

### CODE REVIEW

**Comment by dabail10** [comment]
Here are the results of the QC test. Everything passes. This will only change answers when ponds are on.

ice_thickness_derecho_intel_smoke_gx1_44x1_medium_qc qc_base
ice_thickness_derecho_intel_smoke_gx1_44x1_medium_qc qc_te_minus_derecho_intel_smoke_gx1_44x1_medium_qc qc_base
ice_thickness_derecho_intel_smoke_gx1_44x1_medium_qc qc_te

File: columnphysics/icepack_therm_mushy.F90 [hunk 1]
```diff
@@ -3150,7 +3157,11 @@ subroutine flushing_velocity(zTin,   phi,   &
        w = (perm_harm * rhow * gravit * (dhhead / hin)) / viscosity_dyn
 
        ! maximum down flow to drain pond
-       w_down_max = (hpond * apond) / dt
+       if (tr_pond_lvl) then
+          w_down_max = (hpond * apnd * alvl) / dt
+       else
+         w_down_max = (hpond * apnd) / dt
```
apcraig, line 3163: fix indentation

columnphysics/icepack_therm_vertical.F90
```diff
@@ -107,7 +107,7 @@ subroutine thermo_vertical (dt,          aicen,     &
                                   congel,      snoice,    &
                                   mlt_onset,   frz_onset, &
                                   yday,        dsnow,     &
-                                  prescribed_ice)
+                                  prescribed_ice, alvl)
```
apcraig, line 110: Suggest moving alvl up in the calling arguments.  Makes no difference technically, but seems like it belongs with apnd and hpond, for instance.

davidclemenssewall: Done.

apcraig:
Is this ready to merge?  Is any other discussion needed?

davidclemenssewall:
I believe this is ready to merge but please let me know if further input is needed.

eclare108213:
I'm sorry I haven't had a chance to look at this carefully until now.  I agree with the change from `rhow` to `rhowfresh`. Thanks for noticing that.  

The naming convention for ponds is that `apnd` is the tracer and `apond` is the area fraction, with `apond=apnd*alvl` for the level-pond scheme and `apond=apnd` for the others, as you point out.  I disagree with sending both `apnd` and `alvl` into the physics routines and then using if blocks to distinguish among the pond schemes.  We should be sending the same physical quantity (pond fraction) into the physics routines -- `apond` is (or should be) used in all of those.  Also, `hpond` and `apond` should be computed and used consistently throughout the code.  

There might still be a bug, e.g. if we aren't actually sending in `apond` when needed.  But be careful with whether the fraction is of the grid cell area or of the ice.  The variable names might not be consistently defined.

I think you already understand the following -- I'm writing it out to remind myself how it works and to make sure we are on the same page:

The thermo routines act on whole categories of ice thickness, i.e. the full ice area fraction covered by each thickness category; the category column is considered uniform (a single point in space) for purposes of the vertical thermo calculation.  A fraction of the category area is ridged; a fraction is level; a fraction is snow-covered; a fraction has melt ponds on it, etc.  Since the thermodynamics calculation is for the whole category area, we average over the relevant fractions as needed.  

The area tracers are carried on area fractions of things.  For topo ponds, the melt pond tracer `apndn` is the fraction of the ice in a category; `sum(apndn*aicen)` over n categories is the pond fraction of the full grid cell area.  Level-ice ponds are carried on only the level portion of the ice; in that case the tracer `apndn` is the fraction of the level ice and `apndn*alvln` is the fraction of the ice (`apond`).

Your changes are consistent with this, which then prompts the question of which melt pond fraction is being sent in at the top level. If it is wrong, then I would prefer that it be fixed at a higher level, probably in the `icepack_step_*` routines (but check what the icepack and cice drivers are doing).  Is there an issue with doing that instead of modifying the lower-level routines?  I looked back at older versions of the code (9+ years ago) and the subroutine arguments don't seem to have changed functionally, although it's difficult to tell because the code has been refactored and interfaces changed a lot since then.  

Re naming conventions:  we try to stick to them, but we usually drop the category n when deep in the (vertical) physics routines.  We continue to clean up from 30 years of coding...

To sum up, we should definitely fix the water density being used and there may well be a bug in the pond area used for level-ice ponds.  After my brief look at the code, I tend to agree with you that there is.  If so, I'd expect the new code to have smaller pond areas and so less pond influence on the ice pack.

davidclemenssewall:
That all makes sense. I have refactored the code to move the calculation of ``apond`` (alvl*apnd for the level ice, else apnd) into icepack_step_therm1 and pass apond into the physics routines. While doing this, I discovered a related bug in how the formdrag uses the pond fraction (basically, it uses apnd as if it was apond), I will fix that in the next commit.

columnphysics/icepack_therm_mushy.F90
```diff
@@ -5,12 +5,12 @@ module icepack_therm_mushy
...
   use icepack_parameters, only: c0, c1, c2, c8, c10
   use icepack_parameters, only: p01, p05, p1, p2, p5, pi, bignum, puny
-  use icepack_parameters, only: viscosity_dyn, rhow, rhoi, rhos, cp_ocn, cp_ice, Lfresh, gravit
+  use icepack_parameters, only: viscosity_dyn, rhow, rhoi, rhos, cp_ocn, cp_ice, Lfresh, gravit, rhofresh
   use icepack_parameters, only: hs_min, snwgrain
   use icepack_parameters, only: a_rapid_mode, Rac_rapid_mode, tscale_pnd_drain
   use icepack_parameters, only: aspect_rapid_mode, dSdt_slow_mode, phi_c_slow_mode
   use icepack_parameters, only: sw_redist, sw_frac, sw_dtemp
-  use icepack_tracers, only: nilyr, nslyr, tr_pond
+  use icepack_tracers, only: nilyr, nslyr, tr_pond, tr_pond_lvl
```
eclare108213, line 13: tr_pond_lvl is no longer needed here

eclare108213 (commented):
Thanks @davidclemenssewall.  As long as you're checking apnd/apond, are you also checking hpnd/hpond to make sure those average depths are correct?

davidclemenssewall:
Most recent commit now passes ``apond`` into ``neutral_drag_coeffs``. I also removed ``hpnd`` and ``ipnd`` from that subroutine since they are not used in it.

columnphysics/icepack_therm_mushy.F90
davidclemenssewall: Fixed.

davidclemenssewall:
> Thanks @davidclemenssewall. As long as you're checking apnd/apond, are you also checking hpnd/hpond to make sure those average depths are correct?

I have not done an exhaustive check, but I have not encountered any errors with hpnd/hpond. Is there a semantic distinction between 'hpnd' and 'hpond'? Within the icepack code itself it looks to me like they are used interchangeably and they always refer to the depth of the **water within the area covered by ponds**. As far as I'm aware it is only in the CICE history driver that we convert into the mean depth of the water within the entire grid cell, averaging together ponded and unponded ice.

eclare108213:
> they always refer to the depth of the **water within the area covered by ponds**.

Good, that makes sense.  The pond depth should be defined as a tracer on the pond area.  Thanks.

columnphysics/icepack_atmo.F90
```diff
@@ -526,9 +525,7 @@ subroutine neutral_drag_coeffs (apnd,     hpnd,     &
       use icepack_tracers, only: tr_pond
 
       real (kind=dbl_kind), dimension (:), intent(in) :: &
-         apnd     ,& ! melt pond fraction of sea ice
-         hpnd     ,& ! mean melt pond depth over sea ice
-         ipnd     ,& ! mean ice pond depth over sea ice in cat n
```
eclare108213, line 531: This isn't going to be backward compatible.  For now, I recommend just adding a comment that they aren't used.

eclare108213: I take that back - this routine is internal to Icepack so it should be fine.  We only worry about backward compatibility for the interface routines.

apcraig: Correct.

apcraig:
I'm running some test suites on derecho now with Icepack and CICE.  If things look OK, I'll approve and merge unless I hear otherwise.  Do we need to rerun QC or do anything else to check results given most recent changes?

columnphysics/icepack_therm_mushy.F90
```diff
@@ -72,15 +72,15 @@ subroutine temperature_changes_salinity(dt,                 &
          lhcoef      , & ! transfer coefficient for latent heat
          Tbot        , & ! ice bottom surfce temperature (deg C)
          sss             ! sea surface salinity (PSU)
-
+
```
eclare108213, line 75: remove added whitespace

columnphysics/icepack_therm_mushy.F90
```diff
@@ -3138,8 +3138,8 @@ subroutine flushing_velocity(zTin,   phi,   &
        perm_harm = real(nilyr,dbl_kind) / perm_harm
 
        ! calculate ocean surface height above bottom of ice
-       hocn = (ice_mass + hpond * apond * rhow + hsn * rhos) / rhow
-
+       hocn = (ice_mass + hpond * apond * rhofresh + hsn * rhos) / rhow
+
```
eclare108213, line 3142: remove added whitespace

columnphysics/icepack_therm_mushy.F90
```diff
@@ -3151,7 +3151,7 @@ subroutine flushing_velocity(zTin,   phi,   &
 
        ! maximum down flow to drain pond
        w_down_max = (hpond * apond) / dt
-
+
```
eclare108213, line 3154: remove added whitespace

eclare108213 (approved):
Approving with just some minor cleanup needed.  I think full testing with the test suites is sufficient, without additional QC runs.  Thank you!

columnphysics/icepack_therm_mushy.F90
davidclemenssewall: done

columnphysics/icepack_therm_mushy.F90
davidclemenssewall: done

columnphysics/icepack_therm_mushy.F90
davidclemenssewall: done

apcraig:
This test fails in CICE, derecho_intel_restart_gbox128_8x1_diag1,

```
 (abort_ice)ABORTED: 
 (abort_ice) called from ice_history_write.F90
 (abort_ice) line number          958
 (abort_ice) error = 
 (ice_check_nc) NetCDF: Numeric conversion not representable, (ice_write_hist) ERROR: writing variable apond

cice               000000000061CD14  ice_gather_scatte         232  ice_gather_scatter.F90
cice               00000000007B48A3  ice_history_write         949  ice_history_write.F90
cice               000000000072C8F7  ice_history_mp_ac        4134  ice_history.F90
cice               0000000000412A15  cice_initmod_mp_c          54  CICE_InitMod.F90
cice               000000000040DDD5  MAIN__                     43  CICE.F90
```

This is writing out the history initial condition file.  It does not fail on startup, just on restart.  And just for this gbox128 case, not for other gx3 cases that I've tried (still more cases to test).  My guess is that apond is being read off the restart and contains NaNs or something.  And that probably happens because some physics is off for this case and the apond restart field had bad data.  Or something like it.  @davidclemenssewall, can you look into this.  I can too if you need me to.  Thanks.

davidclemenssewall:
> This test fails in CICE, derecho_intel_restart_gbox128_8x1_diag1,
> 
> ```
>  (abort_ice)ABORTED: 
>  (abort_ice) called from ice_history_write.F90
>  (abort_ice) line number          958
>  (abort_ice) error = 
>  (ice_check_nc) NetCDF: Numeric conversion not representable, (ice_write_hist) ERROR: writing variable apond
> 
> cice               000000000061CD14  ice_gather_scatte         232  ice_gather_scatter.F90
> cice               00000000007B48A3  ice_history_write         949  ice_history_write.F90
> cice               000000000072C8F7  ice_history_mp_ac        4134  ice_history.F90
> cice               0000000000412A15  cice_initmod_mp_c          54  CICE_InitMod.F90
> cice               000000000040DDD5  MAIN__                     43  CICE.F90
> ```
> 
> This is writing out the history initial condition file. It does not fail on startup, just on restart. And just for this gbox128 case, not for other gx3 cases that I've tried (still more cases to test). My guess is that apond is being read off the restart and contains NaNs or something. And that probably happens because some physics is off for this case and the apond restart field had bad data. Or something like it. @davidclemenssewall, can you look into this. I can too if you need me to. Thanks.

I think that the most likely source of the error is when we convert from apond back to apnd we divide by alvl. Perhaps in that test there are some extremely small values of alvl? I added an if block to zero out apnd if alvl <= puny. Would you mind rerunning the failing test?

apcraig:
Thanks @davidclemenssewall, latest update fixes the problem and works with both debug flags on and off, so that's good.  Will continue testing and keep everyone posted.

apcraig:
Test results look good for Icepack and CICE, nearly all tests have different results as expected.
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#c02d10b31285a3ff4da36ffab6a5c32f2f4ca5fd
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#749e0742c0323cafe1723974b9e5ea4a6cc31bbe
Merging now.

apcraig (approved):

Issue #512, opened 2025-02-04
set_sfcflux arguments are switched in icepack_therm_vertical


kieranricardo:
The `fsurf` and `fcondtop` arguments are switched [here](https://github.com/CICE-Consortium/Icepack/blob/9f4b99afd90416cddc82cfd37ba90eb8ecebded3/columnphysics/icepack_therm_vertical.F90#L2698-L2705), compared to the expected order [here](https://github.com/CICE-Consortium/Icepack/blob/9f4b99afd90416cddc82cfd37ba90eb8ecebded3/columnphysics/icepack_flux.F90#L265). 

Happy to make a quick PR to fix :)

dabail10:
So this code is only used for the Hadley Centre configuration? I guess this is true in one of the Australian comnfigurations as well. Go ahead and submit the PR.

apcraig:
Fixed in #513

PR #513, opened 2025-02-10
Fix sfc_flux argument order in icepack_therm_vertical.F90

kieranricardo:

## PR checklist
    Fixes `set_sfcflux` argument order in icepack_therm_vertical.F90, this caused errors for models using `calc_tsfc=false`. Addresses #512 
- [x] Developer(s): 
   Kieran Ricardo
- [ ] Suggest PR reviewers from list in the column to the right.
    Passed GitHub actions tests: https://github.com/ACCESS-NRI/Icepack/actions/runs/13231902867/job/36930302099?pr=6
    - [x] bit for bit (except for models using `calc_tsfc=false`)
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
- [ ] Please document the changes in detail, including _why_ the changes are made.  This will become part of the PR commit log. 

The call to `set_sfcflux` in `icepack_therm_vertical.F90` has the `fsurfn_f` and `fcondtopn_f` arguments switched, causing issues for models using `calc_Tsfc = .false.`. This PR fixes these.

### DIFF
--- a/columnphysics/icepack_therm_vertical.F90
+++ b/columnphysics/icepack_therm_vertical.F90
@@ -2697,8 +2697,8 @@
                ! hadgem routine sets fluxes to default values in ice-only mode
                call set_sfcflux(aicen      (n),                 &
                                 flatn_f    (n), fsensn_f   (n), &
-                                fcondtopn_f(n),                 &
                                 fsurfn_f   (n),                 &
+                                fcondtopn_f(n),                 &
                                 flatn      (n), fsensn     (n), &
                                 fsurfn     (n),                 &
                                 fcondtopn  (n))


anton-seaice:
This presumably will impact the CICE test that [sets](https://github.com/CICE-Consortium/CICE/blob/a19e5cabcc399e77fb4a99c4c04a279729051f7f/configuration/scripts/options/set_nml.alt03#L12) `calc_Tsfc = .false.`

If it makes sense to run CICE or Icepack tests, I can do them if that is useful @apcraig

dabail10:
We should definitely run the Icepack / CICE tests here to make sure it does not affect configurations with calc_Tsfc = .true. Also, if it is answer changing for calc_Tsfc = .false. then we should do a QC test in CICE.

https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_testing.html#quadratic-skill-validation-test

eclare108213 (approved):
Good catch, thanks.

dabail10 (approved):
I still think it would be useful to see the QC tests here.

apcraig:
I fired off some test suites this morning, waiting to see what they report.  We have no tests for Icepack with calc_tsfc=.false. so just checking whether all tests are bit-for-bit.  We do have a test in CICE with calc_tsfc=.false. so checking results for CICE too.  

My understanding is that this is a bug.  calc_tsfc=.true. should be bit-for-bit.  calc_tsfc=.false. should change some results, but in stand-alone mode, is it just a diagnostic bug?  If true, what would QC tell us?  Happy to do QC if it makes sense.  Are you suggesting QC if answers are not bit-for-bit with calc_tsfc=.true.?

eclare108213:
I'm not sure that our tests are able to evaluate whether calc_Tsfc=.false. would be climate-changing, because I don't think our forcing has the fields necessary to produce results equivalent to calc_Tsfc=.true.

dabail10:
@kieranricardo did you evaluate this bug fix in your coupled system?

anton-seaice:
> @kieranricardo did you evaluate this bug fix in your coupled system?

The bug fixed version is the one we've been using, kieran might have records of what impact this had, but as its a bug we might not have paid too much attention to the old results

anton-seaice:
I am surprised about [set_nml.alt03](https://github.com/CICE-Consortium/CICE/blob/666468075b346e2da536c40dc19d964ef2c1c107/configuration/scripts/options/set_nml.alt03#L12) ? either way, i think it's good to just merge

apcraig:
Yes, the alt03 results were bit-for-bit.  Again, I don't think calc_tsfc affects the stand-alone model, although it might impact some diagnostic fields in the stand-alone model that might be on history files.  But the prognostic state is not affected.  At least that's my interpretation.

apcraig:
For the record, test results are here, 

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#666468075b346e2da536c40dc19d964ef2c1c107
https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#5b0b3fd89dd9c7109bdd487ffdaf8b47c13c7d7e

gbox128 is not bit-for-bit with past weekend results due to merge this week.

PR #518, opened 2025-03-16
Add GEOS coupling capability including heatflux and massflux modifications

apcraig:
## PR checklist
    Add GEOS coupling capability including heatflux and massflux modifications
- [X] Developer(s): 
    zhaobin74, apcraig
    Standalone model is bit-for-bit, testing continues.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Add GEOS heatflux and massflux updates.  This allows coupling to the GEOS
coupled system where a semi-implicit thermodynamic coupling scheme is introduced. Similar to the explicit case, the fields fsurfn are provided by the coupler, along with their derivatives with respect to surface temperature dfsurfn_dTs. In this case, calc_Tsfc is still set to true, allowing ice surface and internal temperature to be updated implicitly. The resultant surface temperature change is passed back to the atmosphere model via coupler to complete the full update of its temperature profiles. This middle-ground approach, enabled by geos_heatflux=true, does not sacrifice accuracy because it does not need limiting effective conductivity as in the explicit case. In addition, in GEOS, the atmsophere
model assumes vapor deposits or sublimates at 0 degC. In this case, mass conservation is enforced and the resulting discrepancy in energy is resolved by another term sblx and passed to ocean. This option is only on when geos_massflux=true.

- Add 4 new shortwave terms, uvrdr, uvrdf, pardr, pardf to the coupling.  These terms represent a breakdown of the direct and diffuse visible shortwave terms into two components, par = photosynthetical active radiation and uvr = rest of the visible shortwave term.  The current visible shortwave is exactly represented by these two components.  Includes adding atm forcing and terms associated with radiation passthru to the ocean.

- Add calculation of GEOS heatflux.  In GEOS, surface and latent heat flux is computed in the atmosphere at 0degC.  The sea ice model has to respect that calculation, but then computes the d(dh)/dTs terms to correct the heatflux for the sea ice temperature which is then applied conservatively in the coupled system. Implementation includes turning off some of the heat flux calculations in Icepack.

- Add calculation of GEOS massflux term.  An equivalent correction is needed to the mass and enthalpy terms to take into account the GEOS coupling.

- Add geos_heatflux and geos_massflux to parameters and namelist input

- Update discover port

- Update documentation

### DIFF
--- a/columnphysics/icepack_parameters.F90
+++ b/columnphysics/icepack_parameters.F90
@@ -155,6 +155,8 @@
          calc_Tsfc     = .true. ,&! if true, calculate surface temperature
                                   ! if false, Tsfc is computed elsewhere and
                                   ! atmos-ice fluxes are provided to CICE
+         geos_heatflux = .false.,&! geos heatflux option
+         geos_massflux = .false.,&! geos massflux option
          update_ocn_f = .false. ,&! include fresh water and salt fluxes for frazil
          solve_zsal   = .false. ,&! if true, update salinity profile from solve_S_dt
          modal_aero   = .false. ,&! if true, use modal aerosal optical properties
@@ -562,7 +564,7 @@
          qqqice_in, TTTice_in, qqqocn_in, TTTocn_in, &
          ktherm_in, conduct_in, fbot_xfer_type_in, calc_Tsfc_in, dts_b_in, &
          update_ocn_f_in, ustar_min_in, hi_min_in, a_rapid_mode_in, &
-         cpl_frazil_in, &
+         cpl_frazil_in, geos_heatflux_in, geos_massflux_in, &
          Rac_rapid_mode_in, aspect_rapid_mode_in, &
          dSdt_slow_mode_in, phi_c_slow_mode_in, &
          phi_i_mushy_in, shortwave_in, albedo_type_in, albsnowi_in, &
@@ -697,6 +699,9 @@
          calc_Tsfc_in    , &! if true, calculate surface temperature
                             ! if false, Tsfc is computed elsewhere and
                             ! atmos-ice fluxes are provided to CICE
+         geos_heatflux_in, &! compute dfsurf/dT, dflat/dT terms instead of fsurf, flat 
+         geos_massflux_in, &! compute mass/enthalpy correction when evaporation/sublimation
+                            ! computed outside at 0C
          update_ocn_f_in    ! include fresh water and salt fluxes for frazil
 
       real (kind=dbl_kind), intent(in), optional :: &
@@ -1166,6 +1171,8 @@
       if (present(conduct_in)           ) conduct          = conduct_in
       if (present(fbot_xfer_type_in)    ) fbot_xfer_type   = fbot_xfer_type_in
       if (present(calc_Tsfc_in)         ) calc_Tsfc        = calc_Tsfc_in
+      if (present(geos_heatflux_in)     ) geos_heatflux    = geos_heatflux_in
+      if (present(geos_massflux_in)     ) geos_massflux    = geos_massflux_in
       if (present(cpl_frazil_in)        ) cpl_frazil       = cpl_frazil_in
       if (present(update_ocn_f_in)      ) update_ocn_f     = update_ocn_f_in
       if (present(dts_b_in)             ) dts_b            = dts_b_in
@@ -1574,8 +1581,8 @@
          Lfresh_out, cprho_out, Cp_out, ustar_min_out, hi_min_out, a_rapid_mode_out, &
          ktherm_out, conduct_out, fbot_xfer_type_out, calc_Tsfc_out, dts_b_out, &
          Rac_rapid_mode_out, aspect_rapid_mode_out, dSdt_slow_mode_out, &
-         phi_c_slow_mode_out, phi_i_mushy_out, shortwave_out, &
-         albedo_type_out, albicev_out, albicei_out, albsnowv_out, &
+         phi_c_slow_mode_out, phi_i_mushy_out, shortwave_out, geos_heatflux_out, &
+         albedo_type_out, albicev_out, albicei_out, albsnowv_out, geos_massflux_out, &
          albsnowi_out, ahmax_out, R_ice_out, R_pnd_out, R_snw_out, dT_mlt_out, &
          rsnw_mlt_out, dEdd_algae_out, &
          kalg_out, R_gC2molC_out, kstrength_out, krdg_partic_out, krdg_redist_out, mu_rdg_out, &
@@ -1715,6 +1722,9 @@
          calc_Tsfc_out    ,&! if true, calculate surface temperature
                             ! if false, Tsfc is computed elsewhere and
                             ! atmos-ice fluxes are provided to CICE
+         geos_heatflux_out,&! compute dfsurf/dT, dflat/dT terms instead of fsurf, flat 
+         geos_massflux_out,&! compute mass/enthalpy correction when evaporation/sublimation
+                            ! computed outside at 0C
          update_ocn_f_out   ! include fresh water and salt fluxes for frazil
 
       real (kind=dbl_kind), intent(out), optional :: &
@@ -2218,6 +2228,8 @@
       if (present(conduct_out)           ) conduct_out      = conduct
       if (present(fbot_xfer_type_out)    ) fbot_xfer_type_out = fbot_xfer_type
       if (present(calc_Tsfc_out)         ) calc_Tsfc_out    = calc_Tsfc
+      if (present(geos_heatflux_out)     ) geos_heatflux_out= geos_heatflux
+      if (present(geos_massflux_out)     ) geos_massflux_out= geos_massflux
       if (present(cpl_frazil_out)        ) cpl_frazil_out   = cpl_frazil
       if (present(update_ocn_f_out)      ) update_ocn_f_out = update_ocn_f
       if (present(dts_b_out)             ) dts_b_out        = dts_b
@@ -2528,6 +2540,8 @@
         write(iounit,*) "  conduct    = ", trim(conduct)
         write(iounit,*) "  fbot_xfer_type = ", trim(fbot_xfer_type)
         write(iounit,*) "  calc_Tsfc  = ", calc_Tsfc
+        write(iounit,*) "  geos_heatflux = ", geos_heatflux
+        write(iounit,*) "  geos_massflux = ", geos_massflux
         write(iounit,*) "  cpl_frazil = ", cpl_frazil
         write(iounit,*) "  update_ocn_f = ", update_ocn_f
         write(iounit,*) "  dts_b      = ", dts_b
--- a/columnphysics/icepack_therm_itd.F90
+++ b/columnphysics/icepack_therm_itd.F90
@@ -29,7 +29,7 @@
       use icepack_parameters, only: phi_init, dsin0_frazil, salt_loss
       use icepack_parameters, only: Tliquidus_max
       use icepack_parameters, only: rhosi, conserv_check, rhosmin, snwredist
-      use icepack_parameters, only: kitd, ktherm
+      use icepack_parameters, only: kitd, ktherm, geos_heatflux
       use icepack_parameters, only: z_tracers, hfrazilmin, hi_min
       use icepack_parameters, only: cpl_frazil, update_ocn_f, saltflux_option
       use icepack_parameters, only: icepack_chkoptargflag
@@ -2016,7 +2016,9 @@
       ! Let rain drain through to the ocean.
       !-----------------------------------------------------------------
 
-      fresh  = fresh + frain * aice
+      if (.not.geos_heatflux) then  ! include rain drainage
+         fresh  = fresh + frain * aice
+      endif
 
       !-----------------------------------------------------------------
       ! Given thermodynamic growth rates, transport ice between
--- a/columnphysics/icepack_therm_vertical.F90
+++ b/columnphysics/icepack_therm_vertical.F90
@@ -30,7 +30,7 @@
       use icepack_parameters, only: ustar_min, fbot_xfer_type, formdrag, calc_strair
       use icepack_parameters, only: rfracmin, rfracmax, dpscale, frzpnd, snwgrain, snwlvlfac
       use icepack_parameters, only: phi_i_mushy, floeshape, floediam, use_smliq_pnd, snwredist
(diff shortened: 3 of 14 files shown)


doc/source/science_guide/sg_boundary_forcing.rst
```diff
@@ -88,7 +88,16 @@ stable and accurate procedure would be to compute the temperature
...
 however, given that the atmosphere and sea ice models generally exist on	
-different grids and/or processor sets.
+different grids and/or processor sets. In NASA GMAO GEOS-ESM coupled model, 
+a semi-implicit thermodynamic coupling scheme is introduced. Similar to the explicit
+case, the fields ``fsurfn`` are provided by the coupler, along with their derivatives
+with respect to surface temperature ``dfsurfn_dTs``. In this case, ``calc_Tsfc``
+is still set to true, allowing ice surface and internal temperature to be updated
+implicitly. The resultant surface temperature change is passed back to the
+atmosphere model via coupler to complete the full update of its temperature profiles.
+This middle-ground approach, enabled by ``enforcing_heatflux=true``, does not sacrifice accuracy because it does not need limiting effective conductivity as in the explicit case.
```
eclare108213, line 98: change the wording slightly, to "...it does not need effective conductivity limiting as in the explicit case."

doc/source/science_guide/sg_thermo.rst
```diff
@@ -1802,7 +1802,12 @@ ice), and :math:`L_v = 2.501 \times 10^6 \ \mathrm{J/kg}` is the latent
 heat of vaporization of liquid water at :math:`0^{\circ}C`. Note that :math:`\rho L_v` is
 nearly an order of magnitude larger than typical values of :math:`q`.
 For positive latent heat fluxes, the deposited snow or ice is assumed to
-have the same enthalpy as the existing surface layer.
+have the same enthalpy as the existing surface layer. Some climate models (for 
+example, GEOS-ESM) compute mass flux (sublimation or deposition) in the atmsophere 
+model which assumes vapor deposits or sublimates at 0 degC. In this case, mass
+conservation is enforced and the resulting discrepancy in energy is resolved by
+another term ``sblx`` and passed to ocean. This option is only on when 
+``enforcing_massflux = true``.
```
eclare108213, line 1810: The enforcing_massflux flag probably needs to be more specific, since this is only for this particular type of atmo/ice coupling.

columnphysics/icepack_parameters.F90
```diff
@@ -155,6 +155,8 @@ module icepack_parameters
          calc_Tsfc     = .true. ,&! if true, calculate surface temperature
                                   ! if false, Tsfc is computed elsewhere and
                                   ! atmos-ice fluxes are provided to CICE
+         geos_heatflux = .false.,&! geos heatflux option
+         geos_massflux = .false.,&! geos massflux option
```
eclare108213, line 159: We do need to name these something more descriptive, not using geos.  I'll think about this.
The mass flux is only associated with evaporation and sublimation.  Is that also true for the heat flux here?  Or does it also include contributions from conduction etc?

columnphysics/icepack_therm_vertical.F90
```diff
@@ -1154,7 +1157,8 @@ subroutine thickness_changes (dt,        yday,     &
          wk1         , & ! temporary variable
          zqsnew      , & ! enthalpy of new snow (J m-3)
          hstot       , & ! snow thickness including new snow (m)
-         Tmlts           ! melting temperature (deg C)
+         Tmlts       , & ! melting temperature (deg C)
+         sblx            ! flux error due to cond/sub inconsistency
```
eclare108213, line 1161: what do the letters in the symbol `sblx` represent?  I prefer to use intuitive names for things, even if they are highly abbreviated.

columnphysics/icepack_therm_itd.F90
```diff
@@ -2016,7 +2016,9 @@ subroutine icepack_step_therm2(dt,           hin_max,       &
       ! Let rain drain through to the ocean.
       !-----------------------------------------------------------------
 
-      fresh  = fresh + frain * aice
+      if (.not.geos_heatflux) then  ! include rain drainage
+         fresh  = fresh + frain * aice
+      endif
```
eclare108213, line 2021: What does rain draining from the ice have to do with geos_heatflux?  This drainage is required if the level-ice or (new) sealvl pond schemes are used.  (The topo pond scheme is a bit more complicated.)

columnphysics/icepack_therm_vertical.F90
```diff
@@ -2296,6 +2351,10 @@ subroutine icepack_step_therm1(dt,                        &
          fswthru_vdf , & ! vis dif shortwave penetrating to ocean (W/m^2)
          fswthru_idr , & ! nir dir shortwave penetrating to ocean (W/m^2)
          fswthru_idf , & ! nir dif shortwave penetrating to ocean (W/m^2)
+         fswthru_uvrdr,& ! vis uvr dir shortwave penetrating to ocean (W/m^2)
+         fswthru_uvrdf,& ! vis uvr dif shortwave penetrating to ocean (W/m^2)
+         fswthru_pardr,& ! vis par dir shortwave penetrating to ocean (W/m^2)
+         fswthru_pardf,& ! vis par dif shortwave penetrating to ocean (W/m^2)
```
eclare108213, line 2357: The new radiation parameters need to be explained in the documentation.  What are uvr and par?  Are they associated with particular bands?

eclare108213 (commented):
This is just an initial glance through the code changes, which raised several questions.

Has this new heat/massflux coupling approach been tested in GEOS using both BL99 and mushy thermo?

Have the radiation changes been tested for both ccsm3 and dEdd?

columnphysics/icepack_parameters.F90
zhaobin74: The mass flux is only relevant for evap/sublimation. There is no connection to conduction.

columnphysics/icepack_therm_vertical.F90
zhaobin74: I think ```sblx``` was initially coined to be sublimation flux, i.e., the energy difference introduced by sublimation/condensation. Sorry, it is not a particularly good name. Any suggestions are welcome.

columnphysics/icepack_therm_itd.F90
zhaobin74: This is more of legacy in GEOS coupling. We pass full rain flux directly to ocean, unlike other models doing the split (aice, 1-aice). geos_heatflux flag is just to limit this to GEOS. If this is not considered good practice, we may think about changing it in our coupler.

eclare108213: With this rainwater treatment, then you should never turn on topo ponds (or in the future, any pond scheme that assumes rain water is collected).  I recommend aborting the run if this combination occurs.

columnphysics/icepack_parameters.F90
eclare108213: I'm leaving the previous comment to illustrate my confusion here.  Such parallel nameing implies (to me) that they are mass and heat fluxes associated with the same process.

Apparently (correct me if I'm wrong), the 'massflux' option is about the assumed temperature of water vapor deposition/sublimation, and the 'heatflux' option is entirely different, having to do with the coupling strategy for for calculating the surface temperature.  If so, then perhaps we could change `geos_massflux` to `vapor_flux_correction` or something like that.  But I suggested the same thing for `enforcing_massflux` in sg_thermo.rst.  They are both logicals.  What is the difference between `geos_massflux` and `enforcing_massflux`?  Actually, I don't see `enforcing_massflux` (or `enforcing_heatflux`) in the code, only in the docs, so that might be left over from a previous iteration.

Would it make sense to change `geos_heatflux` to `semi-implicit_Tsfc`?  (I suggested that instead of `enforcing_heatflux`)

eclare108213 (changes requested):
These changes are pretty confusing to me, so I've made some suggestions for names that might help (assuming I understand the code correctly).  These might not be the best choices -- let's discuss.

columnphysics/icepack_therm_itd.F90
zhaobin74: Thanks, @eclare108213. We sure don't want to lose the capability to use the topo pond scheme or others in future. I will consult with my team about how to move forward with it.

columnphysics/icepack_therm_vertical.F90
zhaobin74: 100% agree. Will fix.

columnphysics/icepack_parameters.F90
zhaobin74: @eclare108213, sorry about the ```enforcing_heatflux``` and ```geos_heatflux``` confusion. It is just that the documentation is not synced with the code.

I really like ```semi-implicit_Tsfc``` as the replacement of ```geos_heatflux``` as it captures the nature of GEOS coupling. And ```vapor_flux_correction``` is also appropriate name for its purpose.

We do need these two flags as they are associated with two different processes (all related to coupling).

Thanks for these suggestions.

columnphysics/icepack_therm_itd.F90
apcraig: Do we need another namelist variable to specifically control the fresh water coupling like "couple_rain" which is true by default and false for GEOS?  So rain does not fall on the sea ice but snow does?  It seems like some ice processes will be lost by doing this, right?  Would this be a good time to fix this in GEOS so rain is passed to CICE like snow?  Are there other sea ice models in GEOS that would be impacted by this coupling change?

apcraig:
I just updated the Icepack code, 

-  Rename geos_heatflux to semi_implicit_Tsfc
-  Rename geos_massflux to vapor_flux_correction
-  Rename sblx to de_vapor
-  Add check to not allow semi_implicit_Tsfc with tr_pond_topo
-  Update documentation

I think that takes care of several of the outstanding tasks.  I still have a question about what, if anything, we should do about the rain term in the fresh water coupling.  Please review the changes and let me know if anything else needs to change in code or documentation.

columnphysics/icepack_parameters.F90
apcraig: done

columnphysics/icepack_therm_vertical.F90
apcraig: I have updated the documentation in the code and in the user guide.  done.

apcraig:
I completed a full suite of testing on derecho, tests look fine.  Some answer changes with some compilers, but this seems to be due to compiler optimization.  I tested a few cases manually with the debug flags and those are bit-for-bit.  I think this could be ready to merge after final review.  https://github.com/CICE-Consortium/Test-Results/wiki/icepack_by_hash_forks#be5757df4ffa79e6bd283445892ce5c31c054780, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#c98b212136e9a716c4d89212f5459fcc4ee9453b

eclare108213:
There are still two outstanding questions from my initial code review:

> Has this new heat/massflux coupling approach been tested in GEOS using both BL99 and mushy thermo?
> Have the radiation changes been tested for both ccsm3 and dEdd?

Neither of these should hinder merging this PR, but if the code has not been tested and validated with some of these combinations, then we should abort if it's attempted and make a note in the documentation about what configuration is trusted.

eclare108213:
> @eclare108213, the two coupling flags and radiation changes has all been tested before this latest updates by @apcraig.

@zhaobin74 sorry to keep harping on this, but let me ask for just a bit more clarification.  

1. What is the standard thermodynamics configuration that you run with the coupling flags and radiation changes?  Are you using mushy thermo or BL99 thermo?  Have you tested your coupled code using the other one?
2. Likewise for the radiation changes, do you use the ccsm3 or dEdd radiation in your standard configuration?  Have you tested the other one?  

It looks like the code has changed for all of these options, and I want to make sure that they have all been tested.

apcraig:
@zhaobin74, do you feel these Icepack and CICE changes are ready to merge?  If so, please "review" the code and approve.  @eclare108213, could you also do a review and see if there are any outstanding issues that need to be addressed and approve if not.  I'm not pushing for this to be merged, just want to keep things moving and if it's ready, we should merge.  Thanks!

zhaobin74:
> @zhaobin74, do you feel these Icepack and CICE changes are ready to merge? If so, please "review" the code and approve. @eclare108213, could you also do a review and see if there are any outstanding issues that need to be addressed and approve if not. I'm not pushing for this to be merged, just want to keep things moving and if it's ready, we should merge. Thanks!

@apcraig, sorry for the delay. I had another walk through of the code and think it is ready. Thanks.

eclare108213 (approved):
@zhaobin74 thanks for testing BL99/Mushy and ccsm3/dEdd.

doc/source/icepack_index.rst
```diff
@@ -190,11 +190,15 @@ section :ref:`tabnamelist`.
    "fswint", "shortwave absorbed in ice interior", "W/m\ :math:`^2`"
    "fswpenl", "shortwave penetrating through ice layers", "W/m\ :math:`^2`"
    "fswthru", "shortwave penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_ai", "grid-box-mean shortwave penetrating to ocean (fswthru)", "W/m\ :math:`^2`"
+   "fswthru_idf", "near IR diffuse shortwave penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_idr", "near IR direct shortwave penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_pardf", "visible photosynthetically active shortwave diffuse radiation 400-700nm penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_pardr", "visible photosynthetically active shortwave direct radiation 400-700nm penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_uvrdf", "visible ultraviolet shortwave diffuse radiation > 700nm penetrating to ocean", "W/m\ :math:`^2`"
+   "fswthru_uvrdr", "visible ultraviolet shortwave direct radiation > 7000nm penetrating to ocean", "W/m\ :math:`^2`"
```
eclare108213, line 199: should probably be 700 nm rather than 7000

eclare108213:
Sorry I wasn't fast enough -- looked back while reviewing the cice PR and saw a couple of other things.  Not super important and can be corrected later (but soon, before we forget!).

PR #542, opened 2025-09-19
Fix Tsnice calculation

dabail10:

## PR checklist
Bug fix for Tsnice computation. Addresses part of CICE Issue https://github.com/CICE-Consortium/CICE/issues/1033. More coming in CICE PR.
- [X] Developer(s): 
dabail10 (D. Bailey)
 Test results coming. 
    - [X] bit for bit
    - [ ] more substantial 
    - [X] Yes
    - [ ] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
This should be a bfb change for the model. This will change the output of the CICE history variable f_sitempsnic.

### DIFF
--- a/columnphysics/icepack_therm_vertical.F90
+++ b/columnphysics/icepack_therm_vertical.F90
@@ -378,9 +378,9 @@
       Tsnice = c0
       if ((hslyr+hilyr) > puny) then
          if (hslyr > puny) then
-            Tsnice = (hslyr*zTsn(nslyr) + hilyr*zTin(1)) / (hslyr+hilyr)
+            Tsnice = Tsnice + aicen*((hslyr*zTsn(nslyr) + hilyr*zTin(1)) / (hslyr+hilyr))
          else
-            Tsnice = Tsf
+            Tsnice = Tsnice + aicen*Tsf
          endif
       endif
 


columnphysics/icepack_therm_vertical.F90
```diff
@@ -378,9 +378,9 @@ subroutine thermo_vertical (dt,          aicen,     &
       Tsnice = c0
       if ((hslyr+hilyr) > puny) then
          if (hslyr > puny) then
-            Tsnice = (hslyr*zTsn(nslyr) + hilyr*zTin(1)) / (hslyr+hilyr)
+            Tsnice = Tsnice + aicen*((hslyr*zTsn(nslyr) + hilyr*zTin(1)) / (hslyr+hilyr))
```
NickSzapiro-NOAA, line 381: Can I check that I understand? Does this aicen factor get "canceled" in ice_history in:
```
! Only average for timesteps when ice present
if (avail_hist_fields(n)%avg_ice_present) then
   do j = jlo, jhi
   do i = ilo, ihi
      if (tmask(i,j,iblk)) then
            a2D(i,j,n,iblk) = &
            a2D(i,j,n,iblk)*avgct(ns)*ravgip(i,j)
      endif
```

dabail10: No, this is aggregating the category Tsnice into the grid cell Tsnice. The time averaging in ice_history.F90 will come in a CICE PR!

dabail10:
> The range of sitempsnic is certainly better in 24-hour history from ufs-weather-model cpld_control_gfsv17 RT, especially no (huge) negatives. Left is current, right is this PR (with different colorbars): image

Great. There is another fix coming in CICE to account for a case when aice_init is zero, but aice is not.

columnphysics/icepack_therm_vertical.F90
anton-seaice: Oh - maybe this line makes would be correct now:



`worka(i,j) = aice(i,j,iblk)*(Tsnice(i,j,iblk)/aice_init(i,j,iblk)+Tffresh)`

https://github.com/CICE-Consortium/CICE/blob/b01032506d71d86ca51cef8c3c28c1ab87532833/cicecore/cicedyn/analysis/ice_history.F90#L2768

the divide by aice_init corrects it for the aice weighting applied during thermo vertical

columnphysics/icepack_therm_vertical.F90
```diff
@@ -378,9 +378,9 @@ subroutine thermo_vertical (dt,          aicen,     &
       Tsnice = c0
       if ((hslyr+hilyr) > puny) then
          if (hslyr > puny) then
```
eclare108213, line 380: If `hslyr > puny`, then `hslyr+hilyr > puny` since `hilyr >= 0`, so then is the outer conditional necessary?  It seems redundant, logically.

dabail10: Good catch.

columnphysics/icepack_therm_vertical.F90
```diff
@@ -378,9 +378,9 @@ subroutine thermo_vertical (dt,          aicen,     &
       Tsnice = c0
```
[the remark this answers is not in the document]
dabail10: Tsnice is now initialized in icedrv_flux.F90 and ice_flux.F90.

apcraig:
As part of these updates, it makes sense to me to do some manual checks of the updated diagnostic output.  Maybe run the model for a day, write out a bunch of data either in history or via stdout, and verify that the new diagnostics are generating results consistent with the raw data.  If we do that, these errors should get caught, and we'll have more confidence in the results as well.

apcraig (commented):
Needs further review

dabail10:
@anton-seaice @eclare108213 @NickSzapiro-NOAA Have I addressed all of your concerns here? I know Nick is not able to look at this right now.

dabail10:
Also, I did a search again in case there was already an interface temperature. There is iTin in the icepack_brine.F90 module, but I couldn't find one otherwise.

eclare108213:
@dabail10 I don't think your thickness-weighted temperature formula gives you the temperature at the interface.  That is probably the average temperature over the two layers, which isn't the same thing as the interface temperature. Setting up a grid with temperature on the y axis and distance on the x axis (i.e. the snow-ice column lying sideways), I get the interface temperature as -7.42857C in your example above with -10C, 0.1 m snow and -1C, 0.25 m ice layers.

dabail10:
I agree. I had naively assumed it was a weighted average of the two layers. Instead maybe the straight line fit of:

T = (Tice - Tsno) / 0.5 *(hslyr + hilyr) * x + Tsno

So, when x = 0, T = Tsno. when x = 0.5*(hslyr+hilyr), T=Tice. With this in mind, x = 0.5*hslyr is the snow-ice interface and T = (Tice-Tsno)*hslyr / (hilyr+hslyr) + Tsno.

Using Tsno = -10 and Tice = -1, hslyr=0.1, hilyr=0.25, this gives -7.429 as you found.

dabail10:
> > T = (Tice-Tsno)*hslyr / (hilyr+hslyr) + Tsno
> 
> reduces to
> 
> `T = (Tice * hslyr + Tsno * hilyr) / (hslyr + hilyr)`
> 
> Same number of flops but the interpolant is a little easier to recognize.

Which is funny, because I was wondering if I was doing the weighted average correctly. This expression weights the snow temperature with hilyr, whereas my version was weighting it with hslyr.

PR #543, opened 2025-09-25
Zap residual ice

eclare108213:
Adds minimum area and mass parameters and zaps grid-cell volumes of ice less than those.
- [x] Developer(s): 
@eclare108213, @DeniseWorthen 
Significant testing is documented in the comments below, including Icepack test suites, QC tests in CICE and ice-ocean coupled simulations.  The results are not BFB but do change the results in regions where there was little ice to begin with.  QC tests indicate these are not climate-changing differences.
    - [ ] bit for bit
    - [x] more substantial 
    - [x] Yes
    - [ ] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 
This PR addresses the long-standing residual ice problem in CICE by zapping grid-cell volumes with area or mass less than the new namelist parameters `min_area` and `min_mass`.  To work properly, these parameters should likely be set to the same values as the dynamics area and mass minimum values `dyn_area_min` and `dyn_area_mass`.  Since those parameters in CICE are quite large (and everyone should decrease them), the default values in Icepack are currently set to 1.e-11 and 1.e-10, respectively, maintaining the area/mass=0.1 relationship of the dynamics parameters. The driver (e.g. CICE) should initialize them as needed via an `icepack_init_parameters` call.


DeniseWorthen:
I've got your residual zap branch up and running now and I'm going to do some longer comparison runs w/ our current develop branch and the ICs that start w/ residual ice.

For sensitivity, are there suggested 'non-large' values of ``dyn_area_min`` and ``dyn_area_mass`` that I should try?

eclare108213:
> I've completed 4 runs with our 1/4deg configuration

Yay! This mostly makes sense to me.  After it settles down, the green line indicates about how much ice you have with concentrations between 1.e-7 and 1.e-6.  The larger criteria used for the orange line zaps everything less than 1.e-6 and then some. 

By changing `dyn_area_min` and `dyn_mass_min`, in addition to zapping you're also changing where the dynamics calculation occurs for values less than those in your base case, so it's not surprising that the red line starts off differently.  At least for these few months, it looks like your configuration doesn't crash with the puny ice area (red), but it's interesting that it grows after the initial drop, while the blue baseline stabilizes, like there is something weird going on in the red case. So maybe that's too small.  It would be informative, I think, to see what the velocity values are around the ice edge.  They might be quite big for the smallest criteria.

I think you should choose the parameters you use based on validation against whatever observational data you use.  There might actually be 1.e-7 bits of ice in some grid cell at some time, but it shouldn't matter much for the simulation and I would guess that you can use larger cutoff values to better match your statistics.

Does this all make sense to you?

NickSzapiro-NOAA:
There are some more places with "minimum" thresholds like hi_min=p01 for icepack_therm_itd, aice>p001 in neutral_drag_coeffs in icepack_atmo, dh_min=p001 in icepack_brine. Are these independent of the parameters here?

eclare108213:
@NickSzapiro-NOAA, good catch. We have tried to pull out all the hardwired parameters but clearly missed some of them. 

`hi_min` is a namelist parameter that is intended to maintain stability in the vertical thermodynamics for thin ice (so mainly the thinnest category).  Given an area, it would be related to the mass minimum (which is the total over all categories), but functionally they are pretty independent.  Conceptually, if the ice thickness in a category is less then `hi_min`, then the vertical thermo calculation isn't done, but the ice might still be transported and ridged and end up greater than both `hi_min` and the zapping minima, allowing the thermodynamics to act on it in the next time step. This is the way the code was originally intended to work, and why we spent so much time trying to include `hi_min` in the zapping criteria.

The `p001` criterion for `aice` in `ice_atmo.F90` should not be hardwired.  I think this could track `dyn_area_min` for the wind stress -- it's only used for the form drag parameterization so shouldn't affect configurations without form drag. This change will need to be tested by someone actually using form drag.

@njeffery, the brine height parameterization uses both `dh_min=0.001` and `thinS=0.05`. Could/should those be related to `hi_min` or the dynamics minima, or are they truly independent?

njeffery:
@eclare108213 :  They should be larger  than or equal to the thermodynamic minimum thickness, but not smaller.

njeffery:
Then it shouldn't be making a difference. That's not a problem though. It just means we always calculate brine height changes rather than assuming its the same as ice thickness.

eclare108213:
> Let me know if you want me to run a full test suite on Derecho with the updated implementation in Icepack and CICE, when ready, before we merge.

That would be awesome, thank you. Since this is not BFB, a QC test will also be needed. But I think there also needs to be broader testing by the various modeling centers before this is merged. 

AND I would really like to get opinions on the parameter values. Should we recommend reducing dyn_area_min and dyn_mass_min in the CICE PR? And if so, to what? I'd rather not use the large values we currently have for zapping, but that is how the code is currently set up in that PR. 

Thinking out loud... maybe 2 QC tests could help decide, one with the parameters as currently set in the CICE PR, and a second one with parameters set at the other extreme, as they are currently in Icepack. Compare both test simulations with the present code and with each other. I wouldn't expect either to be climate changing in that test, but who knows? (Caveat: I'm not sure whether the QC configuration has residual ice.)

eclare108213:
Running the QC tests with the current PR will capture the extremes of potential parameters, I think.  The two tests would be
```
A
dyn_area_min = 1.e-3
dyn_mass_min = 1.e-2
```
and
```
B
dyn_area_min = 1.e-11
dyn_mass_min = 1.e-10
```
The baseline run will have the dynamics parameters as in A but no extra zapping.
Thanks!

apcraig:
The QC tests all pass.  The difference plots don't show much of anything.  There are some differences in the Canadian archipelago and something in the southern ocean, but nothing much at the ice edge.  I attach the three difference plots.  zapbase is the current code.  zapqc1 is A above with 0.001 and 0.01 parameter values.  zapqc2 is B above with 1e-11 and 1e-10 values.

ice_thickness_derecho_intel_smoke_gx1_128x1_medium_qc zapbase_minus_derecho_intel_smoke_gx1_128x1_medium_qc zapqc1

ice_thickness_derecho_intel_smoke_gx1_128x1_medium_qc zapbase_minus_derecho_intel_smoke_gx1_128x1_medium_qc zapqc2

ice_thickness_derecho_intel_smoke_gx1_128x1_medium_qc zapqc1_minus_derecho_intel_smoke_gx1_128x1_medium_qc zapqc2

apcraig:
I do not see any cutoff values in the QC test.  The test reads only 'hi' as far as I can tell.  There is a mask generated for gridcells that never have hi > 0.01 but that is not an aice criteria.  I think the QC test considers all gridcells.  It could be that zapping plays little role.  What's interesting is that baseline - A has the most differences.  This suggests that the baseline is operating a lot like B with, I assume, less zapping happening in both vs B.  My conclusion is that the new implementation is zapping more than the current implementation.  I think that's what we expected.  Whether it's doing that mainly at the ice edge seems like an open question though.

eclare108213:
> There is a mask generated for gridcells that never have hi > 0.01 but that is not an aice criteria. 

In cice.t-test.py:

        Calculate the difference and mask the points where the the difference at
        every timestep is 0, or the sea ice thickness for every timestep is < 0.01 meters
        for data_a or data_b

The hi > 0.01 criteria is probably masking out most small-area cells.  Why this is done here is understandable, to prevent comparing cells that have a little bit of ice in one run with cells that never have any in the other run, which probably shouldn't matter for climate.  But we are looking for exactly those cells!  What happens if the criteria value is reduced?  Could it be set to `min(dyn_mass_min/aice)` over the two runs assuming aice=1, or 1.e-10 here? Maybe QC isn't the best test... but it is good to know that there is a hardwired parameter in there.

apcraig:
Right.  It's a cutoff for "every timestep".  That means the test is probably assessing most of the ice edge for most of the year and would only exclude gridcells that only have very small amounts of ice probably just at the peak of the ice season at the far edge.

anton-seaice:
As far as I can tell this works ok.

I ran two experiments with a 25km global ocean + sea ice model forced with a repeat year forcing of a selected year from JRA55do, starting from a restart about ~40 years into an control experiment

- `control' - without this change
- `small_min` : ( `dyn_area_min = 1.e-11, dyn_mass_min = 1.e-10` )
- `large_min` : (`dyn_area_min = 1.e-3, dyn_mass_min = 1.e-2`)

This change appears to zap ice sucessfully:

image

Although some results are counter-intuitive, e.g. the control (without the extra zapping) appears to have more ice in the 1.e-6 to 1e-3 concentration range than the change with the zapping using `small_min`. (Does this mean the extra zapping is preventing some ice growth ?)

image

The impact on Sea Ice Area is small (the change in maximum area is similar, not shown):

image

It appears to deal with residual ice ok:

Figure is cells where ice exists for all of the final year AND final concentration is <1e-6. There is a weird blip near Coronation Island off the end of the Antarctic Peninsula in the small_min run.  

image

Global ocean temperature is marginally cooler in the large_min run:

image

columnphysics/icepack_itd.F90
```diff
@@ -1074,13 +1074,33 @@ subroutine zap_small_areas (dt,                      &
...
+      do n = 1, ncat
+         if ( abs(aicen(n)) <= puny .and. &
+             (abs(aicen(n)) /= c0 .or. abs(vicen(n)) /= c0 .or. abs(vsnon(n)) /= c0)) then
+            zap_category(n) = .true.
+         endif
+      enddo
+
+      zap_residual = .false.
+      if (aice      < max(min_area, puny) .or. &
+          aice*rhoi < max(min_mass, puny)) zap_residual = .true. ! all categories
```
anton-seaice: How about aborting if `min_area` or `min_mass` is less than `puny? Those setings mean the user has tried to do something which the model doesn't support.

e.g. could implementing as  `max(min_area, puny)` leave people confused about why there is no ice with concentration less than puny, greater than min_area

eclare108213:
> To be able to keep current answers and test change, can this come in via a "zap_residual" namelist option?

Yes, good thought.

anton-seaice:
Thanks @eclare108213 

- the extra zapping does cool the ocean slightly. I guess it accelerates the transfer of heat from ocean -> ice -> atmosphere by extracting the heat from the ocean to do the zapping 

> In your plot with areas between 1.e-6 and 1.e-3, what time of year are the main peaks occurring? (Are those plots SH?)

This is global, peaks are march / september - one for each hemisphere

image


> There is a notable adjustment in the first month or two of the low-min test, which I'm not sure how to interpret.

Maybe this (extra low conc cells) is related to the extra points / lower limits for advection ?

columnphysics/icepack_itd.F90
eclare108213: Setting `min_area < puny` means that ice areas less than `puny` would still be zapped, which (if I remember correctly) is the current behavior.  In that case, we could write a warning that areas less than `puny` will still be zapped and offer the namelist flag as an alternative to turn off zapping completely, with no need to abort. I agree that having `zap_residual` on as the default is what we want.

apcraig (approved):
Testing in Icepack and CICE indicates this is ready to merge.

dabail10:
Just an update. I finally did a sensitivity run in CESM with these changes. I think it might have more of an impact.

https://webext.cgd.ucar.edu/BLT1850/b.e30_alpha08b.B1850C_LTso.ne30_t232_wgx3.299_cice/ice/html/infrastructure/index.html

anton-seaice:
Interesting - it might be neatest to make a new issue ?

So the plots show an increase in sea ice area in both hemispheres in a coupled (with interactive atmosphere) model with zap residual ice enabled?

dabail10:
> Interesting - it might be neatest to make a new issue ?
> 
> So the plots show an increase in sea ice area in both hemispheres in a coupled (with interactive atmosphere) model with zap residual ice enabled?

I think this might just be interannual variability. Also, if you look at the spatial plots, the differences are mainly around the ice edge.

columnphysics/icepack_itd.F90
```diff
@@ -1074,13 +1074,33 @@ subroutine zap_small_areas (dt,                      &
...
+      do n = 1, ncat
+         if ( abs(aicen(n)) <= puny .and. &
+             (abs(aicen(n)) /= c0 .or. abs(vicen(n)) /= c0 .or. abs(vsnon(n)) /= c0)) then
+            zap_category(n) = .true.
+         endif
+      enddo
+
+      zap_residual = .false.
+      if (aice      < max(itd_area_min, puny) .or. &
+          aice*rhoi < max(itd_mass_min, puny)) zap_residual = .true. ! all categories
```
NickSzapiro-NOAA, line 1102: should this be `vice*rhoi` (instead of aice*rhoi) @eclare108213 ?

PR #545, opened 2025-12-05
external wave height option

eclare108213:
## PR checklist
Add option for externally generated significant wave height
- [x] Developer(s): 
@erinethomas @eclare108213 
Base_suite is fully BFB with baseline.
Also tested in modified CICE quick_suite, including the 4 FSD tests. All pass, including comparisons with baseline.
    - [x] bit for bit
    - [ ] more substantial 
    - [x] Yes, it will require driver changes and input forcing to access the new feature, but is backward compatibile without those changes
    - [ ] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

A new namelist flag allows significant wave height to be passed into the ice model from a coupler. In addition, this PR
1. moves `wave_spec_height` out of icepack interface argument lists, since it is initialized via `icepack_init_parameters`
2. consolidates multiple wave-spectrum calculations of significant wave height, to reduce duplication of code, used when `wave_height_type = 'internal'`.
 
This PR is the first step in resyncing the CICE Consortium and E3SM versions of Icepack.  It is in draft form until we ensure it runs as expected in E3SM. A few extra commits show up in the list because of differences in histories between the Consortium and E3SM Icepack repositories, going back to the last time they were in sync. Only the wave-height related changes are new, as shown in the file comparison.

### DIFF
--- a/doc/source/user_guide/interfaces.include
+++ b/doc/source/user_guide/interfaces.include
@@ -780,7 +780,7 @@ icepack_init_parameters
            stefan_boltzmann_in, ice_ref_salinity_in, &
            Tffresh_in, Lsub_in, Lvap_in, Timelt_in, Tsmelt_in, &
            iceruf_in, Cf_in, Pstar_in, Cstar_in, kappav_in, &
-           kice_in, ksno_in, &
+           kice_in, ksno_in, itd_area_min_in, itd_mass_min_in, &
            zref_in, hs_min_in, snowpatch_in, rhosi_in, sk_l_in, &
            saltmax_in, phi_init_in, min_salin_in, Tliquidus_max_in, &
            min_bgc_in, dSin0_frazil_in, hi_ssl_in, hs_ssl_in, hs_ssl_min_in, &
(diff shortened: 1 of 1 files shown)


eclare108213:
I'm thinking of changing the `wave_height_type = 'internal'` option to `'spectral'`, and `wave_height` to `wave_sig_ht`, wherever it appears in the code.  
CICE draft PR will be posted soon to illustrate the changes needed in the driver.
FYI @lettie-roach

columnphysics/icepack_fsd.F90
```diff
@@ -620,7 +620,6 @@ subroutine fsd_add_new_ice (n, &
                                   wave_sig_ht,               &
                                   wave_spectrum,             &
                                   wavefreq,                  &
-                                  dwavefreq,                 &
```
dabail10: I guess I am not sure why you are removing dwavefreq from the call, but not wavefreq?

dabail10 (commented):
I guess this looks fine. I am still confused about why dwavefreq goes away.

lettie-roach:
> > Then we might keep 'wave_spec_type' for the other option (instead of wave_height_type) - as the input data is the wave energy spectrum as a function of frequency, E(f)
> 
> This would be really confusing, I think. We don't have any new wave data for CICE at the moment, but the input data for `wave_height_type = 'coupled'` in E3SM is actually significant wave height calculated by WW3. This PR is intended to support that coupling approach while still allowing the existing options.

Ah, I hadn't realized that E3SM is using wave significant height rather than the spectrum. Then 'wave_height_type' makes sense.

I do think 'wave_frac_method' or 'wave_frac_type' would be less confusing than 'wave_spec_type' for the option to do convergence over random sea surface height phase vs one iteration and constant phase in the wave fracture scheme.

erinethomas:
The wave-ice parts of this PR has now been fully tested in E3SM. I fully approve of all @eclare108213s consolidations!

apcraig (approved):

dabail10 (commented):
Looks good to me.

Issue #556, opened 2026-03-17
Add conductive flux limiting for vertical thermodynamics when `calc_Tsfc=.false.`


kieranricardo:
When the vertical thermodynamic solver is forced by prescribed surface fluxes (`calc_Tsfc = .false.`), it can become numerically unstable. This instability appears to arise in the vertical thermodynamic solve when the conductive flux is too large (which depends on ice thickness and time step). Flux limiting is needed to stabilise the solution in this configuration.

An existing implementation of this flux limiting is available in an ACCESS fork of CICE5. Incorporating a similar approach here would enable stable coupling with the UM.

The conductive flux limiting occurs both before the thermodynamic solve if the conductive flux to ice thickness ratio exceeds a certain value, and during the thermodynamic solve if any of the layer temperatures exceed 0C.

eclare108213:
Is this problem specific to either mushy or BL99 thermodynamics?  It sounds like it is related to a problem from a while back, in which an entire layer within the ice would melt completely within a time step.  That was remedied by using some of the interior forcing to melt the surface ice instead of the interior.  Could that approach work in this case?

The PR linked above has been closed.  Did you find a better way to handle this issue?

kieranricardo:
I've moved from that draft PR to this one: https://github.com/CICE-Consortium/Icepack/pull/557 - sorry for the confusion!

This is a problem with forcing the vertical thermo solver with a prescribed `fcondtop` (`calc_tsfc = .false.`). AFAIK mushy doesn't support `calc_tsfc = .false.` yet, but if it does it might have a similar problem.

Are you referring to the `einex` energy that's added to `fcondbot`? This is pretty similar to that, except it reduces `fcondtop`, and this reduction accumulates over iterations of the solver.

PR #557, opened 2026-03-17
Add conductive flux limiting for vertical thermodynamics when `calc_Tsfc=.false.`

kieranricardo:
## PR checklist
    This limits the conductive flux for the vertical thermodynamic solve when `calc_Tsfc=.false.` See: #556 
- [x] Developer(s): 
    @kieranricardo @blimlim @anton-seaice 
Icepack and CICE tests look good.  Only alt03 CICE tests are not bit-for-bit as expected.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#5b972c241a5865babf262d5a48aefd24320a08cd
    - [X] bit for bit except when calc_tsfc=.false.
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

When the vertical thermodynamic solver is forced by prescribed surface fluxes (calc_Tsfc = .false.), it can become numerically unstable. This instability appears to arise in the vertical thermodynamic solve when the conductive flux is too large (which depends on ice thickness and time step). Flux limiting is needed to stabilise the solution in this configuration.

An existing implementation of this flux limiting is available in an ACCESS fork of CICE5. This implements a similar approach, enabling stable coupling with the UM.

This conductive flux limiting occurs in 2 places:

1. Before the thermodynamic solve. If the conductive flux to ice thickness ratio exceeds a certain value (default `1000 w/m^2`) the conductive flux is reduced such the ratio matches this value. The excess energy is used for bottom melt. 

2. During the thermodynamic solve. If any of the layer temperatures exceed 0C, the layer temperature is reduced to 0C. The excess energy is deducted from the conductive flux and passed to the ocean.

Test results are bit-for-bit except when calc_tsfc=.false. in CICE alt03 tests.


anton-seaice:
This limiting is noted in the icepack docs here:

https://github.com/CICE-Consortium/Icepack/blob/222ab41572cc57e3cd9ee36e16d16cf9223c7f19/doc/source/science_guide/sg_boundary_forcing.rst?plain=1#L83-L85

I can run icepack standalone tests when this is ready

apcraig:
Is this change bit-for-bit with calc_Tsfc=.false.?  Is that because flux limiting is rarely triggered?  When it is triggered, it changes answers, correct?  We probably should run this update in CICE as well as Icepack.

columnphysics/icepack_parameters.F90
```diff
@@ -136,7 +136,9 @@ module icepack_parameters
          aspect_rapid_mode =     1.0_dbl_kind,&! aspect ratio (larger is wider)
          dSdt_slow_mode    = -1.5e-7_dbl_kind,&! slow mode drainage strength (m s-1 K-1)
          phi_c_slow_mode   =    0.05_dbl_kind,&! critical liquid fraction porosity cutoff
-         phi_i_mushy       =    0.85_dbl_kind  ! liquid fraction of congelation ice
+         phi_i_mushy       =    0.85_dbl_kind,&! liquid fraction of congelation ice
+         ratio_Wm2_m       =           1000.0,&! max condutive flux/depth ratio (W m)
+         cold_temp_flag    =        c0 - 60.0  ! min temp used to limit the conductive flux (C)
```
anton-seaice:
```suggestion
         ratio_Wm2_m       =        1000.0_dbl_kind ,&! max condutive flux/depth ratio (W m)
         cold_temp_flag    =        -60.0_dbl_kind  ! min temp used to limit the conductive flux (C)
```
Could you fix alignment here too

columnphysics/icepack_parameters.F90
```diff
@@ -2575,6 +2586,8 @@ subroutine icepack_write_parameters(iounit)
         write(iounit,*) "  dSdt_slow_mode = ", dSdt_slow_mode
         write(iounit,*) "  phi_c_slow_mode = ", phi_c_slow_mode
         write(iounit,*) "  phi_i_mushy= ", phi_i_mushy
+        write(iounit,*) "  ratio_Wm2_m= ", ratio_Wm2_m
+        write(iounit,*) "  cold_temp_flag = ", cold_temp_flag
```
anton-seaice, line 2590: should these be within a `if ( .not. calc_Tsfc)` ?

columnphysics/icepack_therm_vertical.F90
```diff
@@ -2053,6 +2120,10 @@ subroutine conservation_check_vthermo(dt,                 &
          call icepack_warnings_add(warnstr)
          write(warnstr,*) subname, 'Input energy =', einp
          call icepack_warnings_add(warnstr)
+         write(warnstr,*) subname, 'Numerical energy =', e_num
+         call icepack_warnings_add(warnstr)
+         write(warnstr,*) subname, 'fcondtopn_extra energy =', fcondtopn_extra
+         call icepack_warnings_add(warnstr)
```
anton-seaice: Should these have an `if ( .not. calc_Tsfc) `?

anton-seaice:
> Is this change bit-for-bit with calc_Tsfc=.false.? Is that because flux limiting is rarely triggered? When it is triggered, it changes answers, correct? We probably should run this update in CICE as well as Icepack.

Looks like alt03 sets calc_Tsfc=.false., so quite possibly could be answer changing in that test.

https://github.com/CICE-Consortium/CICE/blob/783a867d24520403cb35cec5f07cda8b275448b1/configuration/scripts/options/set_nml.alt03#L12

kieranricardo:
@apcraig flux limiting only happens if `calc_Tsfc=.false.`, so this shouldn't be answer changing for `calc_Tsfc=.true.`

columnphysics/icepack_therm_vertical.F90
kieranricardo: yep this definitely should, I'll add it in

columnphysics/icepack_parameters.F90
kieranricardo: It looks like there's no other `if` checks in `icepack_write_parameters`, so I think all parameters are output regardless of whether they're used. It might be inconsistent to add an if statement here.

columnphysics/icepack_parameters.F90
```diff
@@ -1205,6 +1209,8 @@ subroutine icepack_init_parameters(   &
       if (present(dSdt_slow_mode_in)    ) dSdt_slow_mode   = dSdt_slow_mode_in
       if (present(phi_c_slow_mode_in)   ) phi_c_slow_mode  = phi_c_slow_mode_in
       if (present(phi_i_mushy_in)       ) phi_i_mushy      = phi_i_mushy_in
+      if (present(ratio_Wm2_m_in)       ) ratio_Wm2_m      = ratio_Wm2_m_in
+      if (present(cold_temp_flag_in)    ) cold_temp_flag   = cold_temp_flag_in
```
anton-seaice, line 1213: @kieranricardo has added these as configurable - but we don't plan to add them to the cice namelist (as we dont foresee changing them).

Are we happy to change answers when calc_Tsfc = .false. - @apcraig ? If so, should we not make these two new parameters configurable through `icepack_init_parameters` ?

columnphysics/icepack_parameters.F90
eclare108213: this is okay as long as they are not used elsewhere except when calc_Tsfc = .false.

columnphysics/icepack_therm_bl99.F90
```diff
@@ -122,6 +122,8 @@ subroutine temperature_changes (dt,                 &
          zqsn        , & ! snow layer enthalpy (J m-3)
          zTsn            ! internal snow layer temperatures
 
+      real (kind=dbl_kind), intent(out):: &
+         e_num
```
eclare108213: all local parameters need a comment describing what they are, including units

eclare108213: also fix line spacing here

columnphysics/icepack_therm_bl99.F90
```diff
@@ -558,7 +565,33 @@ subroutine temperature_changes (dt,                 &
                else
                   zTsn(k) = c0
                endif
-               if (l_brine) zTsn(k) = min(zTsn(k), c0)
+               if ((l_brine) .and. zTsn(k)>c0) then
+
+                  if (.not. calc_Tsfc) then
+                     ! Alex West: return this energy to the ocean
```
eclare108213: remove name here and elsewhere -- it can be added at the top of the subroutine with a comment about what the person changed.

eclare108213: Also, please change "this energy" to "excess energy", because you aren't move all of it to the ocean, right?

columnphysics/icepack_therm_bl99.F90
```diff
@@ -683,11 +726,15 @@ subroutine temperature_changes (dt,                 &
...
-            fcondbot = fcondbot + einex/dt
-
-            ferr = abs( (enew-einit)/dt &
+            if (calc_Tsfc) then
+                ! Flux extra energy out of the ice
+                fcondbot = fcondbot + einex/dt
+                ferr = abs( (enew-einit)/dt &
+                     - (fcondtopn - fcondbot + fswint) )
+            else
+                ferr = abs( (enew-einit+e_num)/dt &
```
eclare108213: How are einex and e_num related?  Could you modify einex instead, just above, so that this block of code does not change?

columnphysics/icepack_parameters.F90
eclare108213: Are there values of these parameters that would preserve BFBness when calc_Tsfc = .false.?  If so, I recommend setting the code defaults to those values and allow them to be changed via namelist.

columnphysics/icepack_therm_bl99.F90
eclare108213, line 816: It looks like the 'Matrix ice temperature diff' block (lines 813-820' is redundant.  Not your fault, but please clean this up while you're working in this file.  Thanks!

eclare108213 (changes requested):
Overall, this change seems fine to me.  I would prefer that it be provided as an option whose default behavior does not change answers.  In terms of the coding, it's usually better to use pre-existing variables rather than introducing new ones to do more-or-less the same thing.  I'm not sure how easy it is in this case, but if possible, I recommend trying to use `einex` instead of `e_num` and doing it in a way that minimizes code invasiveness, which might be a bit of a balancing act.

columnphysics/icepack_therm_bl99.F90
kieranricardo: `einex` and `e_num` are both excess energy that's been removed from the ice to bring the temperature below melting. They're calculated slightly differently, `e_num` is accumulated over solver iterations, whereas `einex` isn't. 

The main difference is what's done with this excess energy. `einex` is added to `fcondbot` so it affects bottom melt/growth. And `e_num` is added to `fhocn`, so it's just passed straight to ocean. 

I'm actually not sure what the best approach is for `calc_Tsfc=.false.`, the `einex` approach seems more physical, and would melt additional ice over summer which would be helpful. Thoughts @anton-seaice @ofa001? ACCESS-CM2 did the `e_num` approach for context.

kieranricardo:
Thanks for the review @eclare108213!

Yep `ratio_Wm2_m` and `cold_temp_flag` cold temp are only used when `calc_tsfc=.false.`. Happy to enable these to be changed via a namelist, but what do you mean by BFBness?

columnphysics/icepack_therm_bl99.F90
anton-seaice: Are you suggesting not using e_num at all and seeing if it's stable using the already available einex ? Give it a try I guess, it looks like einex should be calculated fine when calc_Tsfc is false?

eclare108213:
> Thanks for the review @eclare108213!
> 
> Yep `ratio_Wm2_m` and `cold_temp_flag` cold temp are only used when `calc_tsfc=.false.`. Happy to enable these to be changed via a namelist, but what do you mean by BFBness?

BFB = bit-for-bit.  We try to maintain exact reproducibility to the extent possible, except when an answer-changing bug is being fixed. This often requires adding a flag to turn the new feature on and off or to select it.

It would be great if the `einex` functionality could work for your case!

columnphysics/icepack_therm_bl99.F90
kieranricardo: I meant use the current approach for calculating `e_num` (I think we'll need this for stability), but use this for bottom melt/growth instead of passing this to the ocean.

In the code this would add extra checks around einex, calculating it normally if `calc_Tfsc = .true.`, and calculating einex using the e_num approach if `calc_Tfsc = .false.`.

kieranricardo:
@eclare108213 ah gotcha, this will be BFB when `calc_Tfsc=.true.`, but will change answers when it's false.

columnphysics/icepack_parameters.F90
kieranricardo: We could pick extreme value to ensure the `cap_conductive_flux`routine doesn't do anything, but because the flux is also limited in bl99 with `e_num` this will still change answers when `calc_Tsfc = .false.`.

columnphysics/icepack_therm_bl99.F90
kieranricardo: After some discussion it doesn't seem like this approach will work. With `calc_Tfsc=.false.` the extra numerical flux can be much larger than the `calc_Tfsc=.true.`, so adding this to `fcondbot` could cause unrealistic bottom melt/growth. @eclare108213 so it looks like we won't be able to use einex approach here.

eclare108213: Thanks for looking into this.  Then perhaps a compromise will work.  In the PR comments at the top:

> This conductive flux limiting occurs in 2 places:
> 1. Before the thermodynamic solve. If the conductive flux to ice thickness ratio exceeds a certain value (default 1000 w/m^2) the conductive flux is reduced such the ratio matches this value. The excess energy is used for bottom melt.
> 2. During the thermodynamic solve. If any of the layer temperatures exceed 0C, the layer temperature is reduced to 0C. The excess energy is deducted from the conductive flux and passed to the ocean.

Since the excess in 1. is used for bottom melt, that's essentially supplementing `einex`, isn't it?  And then 2. handles it differently.  So perhaps `einex` could be adjusted for 1. instead of using a new variable, and the new variable still used for 2.  If that works, then I would rename the two variables so they are more intuitive, e.g. 1. `einex_melt` and 2. `einex_ocnflux`.  Otherwise, we can keep the current functionality but rename the variables accordingly, e.g. 1. `einex_sfc_calc` and 2. `einex_sfc_flux` or something like that (implicit / explicit?  Will need to look at the GEOS names to try to be consistent).  `einex` translates to 'excess ein' for both cases.

kieranricardo: Good point! But that first approach could be a little involved so I think the second approach sounds good. The first approach would involve adding an extra inout argument to `temperature_changes`, being careful not to update `einex_melt` when `calc_tsfc = .false.`, and would complicate the energy accounting.

kieranricardo: I've updated the variable name to `einex_sfc_calc` and `einex_sfc_flux`, but not sure which GEOS variable names this should match.

columnphysics/icepack_therm_vertical.F90
```diff
@@ -75,6 +76,51 @@ module icepack_therm_vertical
...
+
+         ! AEW: New variables for cold-ice flux capping
+         real (kind=dbl_kind) :: top_layer_temp,     &
+                  reduce_ratio,       &
+                  reduce_amount
+
+
+         if (abs(fcondtopn) > ratio_Wm2_m * hin) then
+            fcondtopn_solve = sign(ratio_Wm2_m * hin,fcondtopn)
+
```
eclare108213: Remove extra blank lines.  Keep them where they help make the code more readable.  In the current version of this subroutine, I suggest removing lines 80, 82, 94, 96, 97 (a comment line), 101, 105.

columnphysics/icepack_therm_vertical.F90
```diff
@@ -75,6 +76,51 @@ module icepack_therm_vertical
...
+         endif
+
+         if ((top_layer_temp < cold_temp_flag) .and. (fcondtopn_solve < c0)) then
+            reduce_ratio = (cold_temp_flag - top_layer_temp) / (c100 + cold_temp_flag)
+            reduce_amount = reduce_ratio * fcondtopn_solve
+            fcondtopn_solve = fcondtopn_solve - reduce_amount
+         endif
+
+      end function cap_conductive_flux
+
```
eclare108213, line 118: Add the standard CICE divider between routines (as in line 78).

columnphysics/icepack_therm_vertical.F90
```diff
@@ -75,6 +76,51 @@ module icepack_therm_vertical
...
+      function cap_conductive_flux(nilyr, nslyr, fcondtopn, hin, zTsn, zTin, hslyr) result(fcondtopn_solve)
+
+         integer (kind=int_kind), intent(in) :: &
+            nilyr   , & ! number of ice layers
+            nslyr       ! number of snow layers
+         real (kind=dbl_kind), intent(in)    :: fcondtopn
+         real (kind=dbl_kind), intent(in)    :: hin
+         real (kind=dbl_kind), intent(in)    :: zTin(nilyr)
+         real (kind=dbl_kind), intent(in)    :: zTsn(nslyr)
```
eclare108213: All declared variables need comments defining them, with units.

eclare108213: elsewhere too, e.g. the new variables in thermo_vertical

columnphysics/icepack_therm_vertical.F90
```diff
@@ -340,6 +393,15 @@ subroutine thermo_vertical (dt,          aicen,     &
 
          else ! ktherm
 
+            if (calc_Tsfc) then
+               fcondtopn_solve = fcondtopn
+            else
+               fcondtopn_solve = cap_conductive_flux(nilyr, nslyr, fcondtopn, hin, zTsn, zTin, hslyr)
+            end if
+
+
```
eclare108213: remove lines 401, 402

eclare108213 (changes requested):
Initial code review.  In addition to the formatting changes noted below, this functionality needs to be documented.  

The GEOS approach is described in sections 2.1 and 2.7 of the Icepack documentation.  It seems to be quite a bit different from what you are doing (especially 2.7).  The last paragraph of section 2.1 generally covers this situation, and I think your approach is now the third one.  This needs to be checked: only the GEOS approach applies to both BL99 and mushy thermo, and the other two only affect BL99.  This needs to be clarified in the docs, along with how the functionality is accessed (e.g. it's on by default in your case).

eclare108213:
@kieranricardo @anton-seaice:  just checking on the status of this PR.  Let me know if you'd like to discuss it.

kieranricardo:
@eclare108213 thanks for your review, I've now added description and units for the all the new variables (and some related existing variables as well).

Shall I edit the documentation in this PR? I can add a description of the limiting in 2.7.4.1.

anton-seaice:
It is covered somewhat in the second paragraph in [2.1,](https://cice-consortium-icepack.readthedocs.io/en/main/science_guide/sg_boundary_forcing.html) so probably update that ?

columnphysics/icepack_therm_bl99.F90
```diff
@@ -763,14 +813,6 @@ subroutine temperature_changes (dt,                 &
          call icepack_warnings_add(warnstr)
          write(warnstr,*) subname, (zTsn(k),k=1,nslyr)
          call icepack_warnings_add(warnstr)
-         write(warnstr,*) subname, 'Matrix ice temperature diff:'
```
eclare108213, line 766: why were these warning statements removed?

kieranricardo:
@eclare108213 @anton-seaice I've updated the documentation in 2.1 and 2.7.4.1. And I think this is ready for another review

doc/source/science_guide/sg_boundary_forcing.rst
```diff
@@ -81,9 +81,8 @@ subroutine, *thermo_vertical*. At the end of the time step, the surface
 temperature and effective conductivity (i.e., thermal conductivity	
 divided by thickness) of the top ice/snow layer in each category are	
 returned to the atmosphere model via the coupler. Since the ice surface	
-temperature is treated explicitly, the effective conductivity may need	
-to be limited to ensure stability. As a result, accuracy may be	
-significantly reduced, especially for thin ice or snow layers. A more	
+temperature is treated explicitly, the conductive flux is limited to ensure stability.
```
eclare108213: To the end of this sentence, please add ", and the excess energy is either applied for bottom melting or fluxed to the ocean."

To the end of the paragraph, add "The GEOS approach is available for either mushy or BL99 thermodynamics, while the conductivity-limiting approaches are only available for BL99."

doc/source/science_guide/sg_thermo.rst
```diff
@@ -1375,9 +1375,15 @@ is that the solution scheme is no longer unconditionally stable.
 Instead, the effective conductivity in the top layer must satisfy a
 diffusive CFL condition:
 
-.. math:: 
+.. math::
    K^* \le {\rho ch \over \Delta t}.
 
+To improve stability, the conductive flux is limited before
```
eclare108213: For clarity:
"To improve stability when ``calc_Tsfc`` = false, the ..."

eclare108213 (changes requested):
Just to be sure I understand, this change will not be backward compatible for anyone running CICE with a hadgem driver or any other driver with calc_Tsfc = false, correct?  Is that okay?  Hadley Centre switched over to the European models a while back so I'm not sure whether it matters at this point, but I'd prefer to keep the code backward compatible in case someone might not want this change.  It does seem like a good improvement, though, maybe even a "bug fix."

columnphysics/icepack_therm_bl99.F90
kieranricardo: typo! added these back in

doc/source/science_guide/sg_boundary_forcing.rst
kieranricardo: Sounds good! 

I've changed the first sentence slightly to: "Since the ice surface	
temperature is treated explicitly, the conductive flux is limited to ensure stability,
with any excess energy either used for bottom melting or transferred to the ocean as a heat flux."

And for the second, I swapped "conductivity-limiting" for "flux-limiting".

kieranricardo:
@eclare108213 thanks for your feedback! Yeah, these changes won't be backwards compatible for `calc_Tsfc=.false.`, although like you say I don't think anyone else is using `calc_Tsfc=.false.`. I did a small bug fix for `calc_Tsfc=.false.` last year, without this icepack crashed with `calc_Tsfc=.false.`, so we can be sure no one was using this up until that point at least.

eclare108213 (approved):
Looks good, thank you.

anton-seaice (approved):
I ran the icepack [base suite](https://github.com/CICE-Consortium/Test-Results/wiki/49b04c6143.gadi.oneapi.26-07-21.030428.0), and got BFB answers except these five tests:


```
FAIL gadi_oneapi_smoke_col_1x1_bgcispol_debug compare ad89637 different-data
FAIL gadi_oneapi_smoke_col_1x1_bgcnice_debug compare ad89637 different-data
FAIL gadi_oneapi_restart_col_1x1_pondlvl compare ad89637 different-data
FAIL gadi_oneapi_restart_col_1x1_bgcispol compare ad89637 different-data
FAIL gadi_oneapi_restart_col_1x1_bgcnice compare ad89637 different-data
```

I guess that is because https://github.com/ACCESS-NRI/Icepack/commit/9c6aaa9cc0c92a107b46998b72c1d14406b08e24 is not in this branch.

anton-seaice:
All tests BFB now using oneapi after updating from main - see https://github.com/CICE-Consortium/Test-Results/wiki/49b04c6143.gadi.oneapi.26-07-21.030428.1

anton-seaice:
Could you review / finalise @apcraig ?

apcraig:
I'll run a CICE test suite and report results.

Just want to confirm that we don't need any new namelist?

anton-seaice:
I don't think we have plans to change / test different values of those parameters, so not required from us.

apcraig:
Testing looks good.  All the icepack tests are bit-for-bit on derecho with intel and gnu.  The CICE tests are also bit-for-bit except the alt03 tests which have calc_tsfc=.false..  I think that is expected.  There is one other test, derecho_gnu_smoke_gbox128_14x2_boxrestore_gridc_reprosum_run10day.fc01, which is also not bit-for-bit for reasons I can't quite figure out.  I think it might have to do with some another recent Icepack PR, but I'm going to ignore.

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#5b972c241a5865babf262d5a48aefd24320a08cd

apcraig (approved):

apcraig:
I added the test results to the PR comments and made a couple other minor updates.  Could @anton-seaice, @kieranricardo review the PR comments and make sure they are up to date given any changes since the PR was created.  You should be edit the PR comments directly.  Once confirmed, I'll merge.  Thanks!

kieranricardo:
Thanks @apcraig, the PR comment looks good to me.

anton-seaice:
Nothing from me either

PR #560, opened 2026-04-03
Update snwredist documentation

### LINKED ISSUE #559: Inconsistency in snwredist option between documentation and code.
The CICE / Icepack documentation says that snwredist = 'snwITDrdg', but in the code it checks snwredist(1:3) = 'ITD'. It turns out that we have not been running with snow redistribution on in CESM3! Do we want the documented string or the code string?

https://cice-consortium-icepack.readthedocs.io/en/main/science_guide/sg_snow.html

apcraig:
## PR checklist
    Update snwredist documentation
- [X] Developer(s): 
    apcraig
    All tests bit-for-bit in Icepack and CICE.  Ran quick and base suites on Derecho on 4 compilers for both standalone Icepack and CICE
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Update snwredist documentation to be consistent with implementation

Add check for snwredist and snw_aging_table valid values

Closes #559


dabail10 (approved):

doc/source/science_guide/sg_snow.rst
```diff
@@ -29,9 +29,7 @@ Snow redistribution
...
-
-
+Three basic approaches are available for snow redistribution by wind, ``snwredist`` = ``bulk``, for which a user-defined parameter :math:`p` (``snwlvlfac``) determines the ratio of snow on ridges to that on level ice, ``snwredist`` = ``ITD``, in which snow can be compacted by the wind or eroded and redeposited on other thickness categories, and ``snwredist`` = ``ITDrdg`` which modifies the ``ITD`` approach to use the standard deviation of the level and ridged ice thickness distribution for snow in suspension to redistribute based on ridged ice area. For all, nonlocal redistribution of snow (i.e., between grid cells) is neglected, assuming that the difference between snow mass blowing into a grid cell and that blowing out is negligible, but snow can be blown into nearby leads and open water.
```
eclare108213: rewrite:
Two basic approaches are available for snow redistribution by wind, ``snwredist`` = ``bulk``, for which a user-defined parameter :math:`p` (``snwlvlfac``) determines the ratio of snow on ridges to that on level ice, ``snwredist`` = ``ITDrdg``, in which snow can be compacted by the wind or eroded and redeposited on other thickness categories based on the standard deviation of the level and ridged ice thickness distribution. For both, nonlocal redistribution of snow (i.e., between grid cells) is neglected, assuming that the difference between snow mass blowing into a grid cell and that blowing out is negligible, but snow can be blown into nearby leads and open water.

eclare108213 (changes requested):
The `ITD` option is not implemented.  It's actually `ITDsd` and currently commented out in the code because we haven't tested it since the initial implementation.  The checks on characters 1:3 are left over from then, and we only offer the ridged-area extended version.  Sorry for the confusion.  In my opinion, it's best to leave it commented out in the code and not mention it in the documentation at all, although it could be commented out there as well.

dabail10:
Another update here. I redid my run with snwredist = 'ITDrdg' and it is not conserving heat. It is also causing the aerosol tracer to not be conserved. I somehow thought this was fixed way back.

apcraig:
OK, I understand the options better now.  I have updated the documentation and checks to support only snwredist=[none,bulk,ITDrdg].  Have another review.  I will update CICE as well once this PR is committed.

If there are problems with energy/conservation with the snow features in general, I wonder if that should be a separate issue/PR at this point?  If there are some fixes/tests that I should do as part of that effort, let me know.  I'm happy to include it in this PR too.

doc/source/science_guide/sg_snow.rst
```diff
@@ -29,9 +29,7 @@ Snow redistribution
...
-
-
+Two basic approaches are available for snow redistribution by wind, ``snwredist`` = ``bulk``, for which a user-defined parameter :math:`p` (``snwlvlfac``) determines the ratio of snow on ridges to that on level ice and ``snwredist`` = ``ITDrdg`` in which snow can be compacted by the wind or eroded and redeposited on other thickness categories based on the standard deviation of the level and ridged ice thickness distribution. For both, nonlocal redistribution of snow (i.e., between grid cells) is neglected, assuming that the difference between snow mass blowing into a grid cell and that blowing out is negligible, but snow can be blown into nearby leads and open water.
```
eclare108213: Please add commas here:  "... on level ice, and ``snwredist`` = ``ITDrdg``, in which ..."

eclare108213 (approved):
Approving with a small editorial change request.  Thank you -

doc/source/science_guide/sg_snow.rst
apcraig: Done.  I will merge this week unless there are other suggestions.  Thanks!

PR #561, opened 2026-04-30
Update Derecho inteloneapi port and fix uninitialized variables

apcraig:
## PR checklist
    Update Derecho inteloneapi port and fix uninitialized variables
- [X] Developer(s): 
    apcraig
    This was tested in standalone Icepack and standalone CICE on derecho with intel, inteloneapi, and gnu with the following results
    - Icepack derecho inteloneapi results are NOT bit-for-bit due to the updated compiler version on derecho with inteloneapi
    - Icepack bgc results are NOT bit-for-bit for all machines and compilers due to fixes in the bgc initialization in the Icepack driver.
    - These Icepack changes applied to CICE are all bit-for-bit in CICE, any bit-for-bit changes in standalone Icepack were implemented in the Icepack driver, not in columnphysics code.

    - [X] bit for bit except derecho inteloneapi and icepack standalone bgc cases
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Update Derecho inteloneapi port.
    
Update the inteloneapi compiler version on derecho to 2025.2.1.
    
Update the intel and inteloneapi compiler debug options on derecho to better trap and warn.  These changes are based on the UFS compiler settings and to help meet UFS requirements.  The new settings are
    
    FFLAGS     += -O0 -check all -check noarg_temp_created -warn -warn noerrors -fp-stack-check -fstack-protector-all -fpe0 -debug -ftrapuv -init=snan,arrays -link_mpi=dbg
    
Fix several issues with uninitialized variables trapped with the new debug flags.  For the most part, standalone Icepack results are bit-for-bit with the code changes except for Icepack bgc configurations because Sil and Nit were not initialized properly.  Moved the init_forcing_bgc call earlier in the initialization calling sequence because sil_data and nit_data are needed in get_forcing_bgc which is called by init_bgc which is called by init_restart early-ish in the initialization sequencing.  get_forcing_bgc is called again later in the initialization sequence and this seems redundant.  It might be worth reviewing the bgc initialization in standalone Icepack.
    
This was tested in standalone Icepack and standalone CICE on derecho with intel, inteloneapi, and gnu with the following results
    - Icepack derecho inteloneapi results are NOT bit-for-bit due to the updated compiler version on derecho with inteloneapi
    - Icepack bgc results are NOT bit-for-bit for all machines and compilers due to fixes in the bgc initialization in the Icepack driver.
    - These Icepack changes applied to CICE are all bit-for-bit in CICE, any bit-for-bit changes in standalone Icepack were implemented in the Icepack driver, not in columnphysics code.


dabail10 (approved):
Great catch on all of the uninitialized variables.

configuration/driver/icedrv_calendar.F90
```diff
@@ -120,6 +120,7 @@ subroutine init_calendar
       istep = 0         ! local timestep number
       time0=istep0*dt   ! start time
       time=istep0*dt    ! s
+      time_forc=c0      ! forcing time, not used anymore
```
eclare108213: If this is not used anymore, can it be completed removed from the code?

apcraig: I wanted to remove it, but it is written to the restart file, so there might be a backwards compatibility issue with regard to those files.  I think we could remove it with relative confidence that there wouldn't be a problem.  I think the only case that could be a problem is using an older version of the code with a newer version of the restart file.  There might also be an issue making sure the binary files are reading the right memory location.  And finally, removing the variable from the restart file would mean regression testing (which compares restart files) would fail with the update, but that's something we can deal with.  This implementation was the least risk, but I'd be happy to remove the variable completely.

eclare108213: I see, it's probably best to just leave it alone for now.  Maybe add to the comment "except in restart files for backward compatibility" or something like that.

eclare108213 (approved):

configuration/driver/icedrv_calendar.F90
apcraig: Good idea, I added some extra comments to the code.

